PR #1, opened 2024-07-10
Update Jumpstarter Architecture Documentation

kirkbrauer:
Improved the formatting and added some clarifications to the Jumpstarter architecture, glossary, and usage.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -23,7 +23,7 @@ unit that can be tested together.
 #### Exporter
 
 The exporter is a service that can run locally or on another Linux device that
-"exports" the interfaces of the Device Under Test (DUT). The exporter
+"exports" the interfaces connected to the Device Under Test (DUT). The exporter
 implements a gRPC service (exposed via a unix socket) that a client can connect
 to directly or through a proxy server to interact with a specific device.
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,53 +1,118 @@
-# Jumpstarter architecture
+# Jumpstarter Architecture
 
-The jumpstarter architecture is constructed around a
-[gRPC protocol](https://github.com/jumpstarter-dev/jumpstarter-protocol/tree/main/proto/jumpstarter/v1) that connects the client to the exporters, either directly, or through a central server.
+The Jumpstarter architecture is based on a client/server model. This enables a
+single client to communicate with one or many devices under test.
 
-The exporters implement a gRPC service (exposed via unix socket), a local client talks directly to this exporter service, and optionally a client that connects to a central server.
+Devices can either be connected to the same machine as the client or distributed
+across remote test runners, for example, in a hybrid cloud CI environment.
 
-The client can connect to a central server that manages the leases and the inventory of exporters and clients, also facilitates routing to the exporter
-gRPC service.
+The core of this architecture is the
+[gRPC protocol](https://github.com/jumpstarter-dev/jumpstarter-protocol/tree/main/proto/jumpstarter/v1)
+that connects a client to a device, either directly, or through a central server.
 
+## Components
 
-## Jumpstarter Router and Controller
+### [`jumpstarter-python`](https://github.com/jumpstarter-dev/jumpstarter-python/)
 
-The jumpstarter-router and controller services live in this repository
-[jumpstarter-router](https://github.com/jumpstarter-dev/jumpstarter-router/).
+#### Device Under Test (DUT)
 
-* `jumpstarter-controller` provides the reconciliation of CRD resources like Exporters
-  and Clients/Identities. Also provides the gRPC to perform exporter leases.
+The DUT is the device that is being tested by Jumpstarter. One or more devices
+can be connected to a single exporter instance so they are treated as a single
+unit that can be tested together.
 
-* `jumpstarter-router` provides the client-to-exporter routing via gRPC streams.
+#### Exporter
+
+The exporter is a service that can run locally or on another Linux device that
+"exports" the interfaces of the Device Under Test (DUT). The exporter
+implements a gRPC service (exposed via a unix socket) that a client can connect
+to directly or through a proxy server to interact with a specific device.
+
+#### Driver
+
+Jumpstarter drivers are exporter plugins that provide the ability to interact with
+a specific device. Drivers provide an interface that can be interacted with
+either through Python (as a library) or the CLI.
+
+#### Client
+
+The Jumpstarter client is a Python library and CLI tool that can connect to one
+or more exporters either locally through socket or remotely through a central server.
+
+#### Tests
+
+Tests are any test scripts written in Python or YAML that utilize the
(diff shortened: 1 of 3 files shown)


docs/architecture.md
```diff
@@ -1,53 +1,118 @@
...
+The DUT is the device that is being tested by Jumpstarter. One or more devices
+can be connected to a single exporter instance so they are treated as a single
+unit that can be tested together.
 
-* `jumpstarter-router` provides the client-to-exporter routing via gRPC streams.
+#### Exporter
 
-## Authentication
+The exporter is a service that can run locally or on another Linux device that
+"exports" the interfaces of the Device Under Test (DUT). The exporter
```
mangelajo, line 26:
```suggestion
"exports" the interfaces connected to the Device Under Test (DUT). The exporter
```

May be this is clearer? :) 

I want to make sure it's clear that we don't run the exporter inside the DUT, but next to it :)

docs/glossary.md
```diff
@@ -1,41 +1,34 @@
...
-  the need for a Jumpstarter service as soon as the interface hardware is available to
-  the local machine.
+* `DUT`: Device Under Test
 
 ## Entities
 
-* `Exporter`: a linux enabled device or a service that exports the interfaces to the devices
-  to be tested. An exporter connects to the Jumpstarter server and waits for commands from
-  a client.
+* `exporter`: A linux service that exports the interfaces of the devices to be tested.
```
mangelajo, line 9:
```suggestion
* `exporter`: A linux service that exports the interfaces to the DUTs.
```

docs/glossary.md
```diff
@@ -1,41 +1,34 @@
...
+  An exporter connects directy to a Jumpstarter server or directly to a client.
 
-* `Client`: a Developer or a CI/CD pipeline that connects to the Jumpstarter server and
-  leases resources from the exporters. The client can run tests on the leased resources.
+* `client`: A Developer or a CI/CD pipeline that connects to the Jumpstarter server
+  and leases resources from the exporters. The client can run tests on the leased
```
mangelajo, line 13:
```suggestion
  and leases exporters. The client can run tests on the leased
```

I would start by the simple model of leasing exporters altogether, then a at a later time I am happy to extend to leasing devices alone if we wanted, but I think that will be more complex.

docs/glossary.md
```diff
@@ -1,41 +1,34 @@
...
 ## Concepts
 
-* `Resource`: a device that is exposed on a exporter, the exporter enumerates those devices
-  and makes them available for use in tests. Examples of resources can be:
-    * network interface
-    * serial port
-    * GPIO pin
-    * Storage device (USB Muxer, SD-Wire, etc.)
-    * a CAN bus interface
+* `Resource`: a device that is exposed on a exporter, the exporter enumerates those
```
mangelajo, line 22:
```suggestion
* `Device`: a device that is exposed on a exporter, the exporter enumerates those
```

Internally we are calling them devices in the gRPC now (I initially wrote resources)

I guess we should change the docs here, or the gRPC

@NickCao  @kirkbrauer  WDYT?

kirkbrauer: Yes, I think a resource is more related to a specific interface or hardware component, so the term device is the lowest unit we should discuss here unless talking about drivers.

docs/usage.md
```diff
@@ -19,48 +20,78 @@ client:
 
 ### Creating a exporter
 
-Creating an exporter registers the CR object in the k8s API, the jumpstarter-controller will create
-an authentication token and attach it to the object.
+To connect a device to Jumpstarter, an exporter instance must be registered.
 
-From the admin point of view it's performed running the following command:
+Exporter creation must be done by an administrator user who has access to
```
mangelajo, line 25: just a note, and why k8s is very cool for this:

It could also be done via gitops with something like ArgoCD or tekton, creating the bare CR (using the CRD) :)   still somebody would need to go and retrieve the token.

docs/glossary.md
NickCao: We are not distinguishing between exported/unexported devices, so `Resource` does not make much sense. Unless we want users to be able to teardown and reinit devices mid test.

docs/usage.md
```diff
@@ -19,48 +20,78 @@ client:
...
 #  --security-opt seccomp=unconfined
 ```
 
-<!-- TODO: create instructions to setup as quadlets with podman and systemd https://www.redhat.com/sysadmin/quadlet-podman -->
+#### Running as a Service
+
+To run the exporter as a service on a test runner with Jumpstarter installed:
+
+```bash
+export JUMPSTARTER_EXPORTER_CONFIG=/etc/jumpstarter/my-exporter.yaml
```
mangelajo, line 76:
```suggestion
export JUMPSTARTER_EXPORTER_CONFIG=/etc/jumpstarter/my-exporter.yaml
```

We would need to set it in the .service file... may be we can abstract the details away by saying that we will do it from the cmdline tool? 

`jumpstarter setup-systemd-exporter /etc/jumpstarter/my-exporter.yaml`

docs/usage.md
```diff
@@ -19,48 +20,78 @@ client:
...
-## User/Client tasks
+When client configuration exists, Jumpstarter will use the specified endpoint
+and token to authenticate with that server
 
-### Authentication/configuration
+#### Configuration
 
-By default the libraries and clients will look for the `~/.jumpstarter/client.yaml` file, this file
-is a client file that contains the endpoint and token to authenticate with the Jumpstarter service.
+By default the libraries and CLI will look for a `~/.jumpstarter/client.yaml`
```
mangelajo, line 92:
```suggestion
By default the libraries and CLI will look for a `~/.config/jumpstarter/client.yaml`
```

So apparently the standard is using .config instead, we had to change it in another project a couple of days ago.

kirkbrauer: Tell that to all the people and projects who don't follow this, my home directory is a mess!

docs/usage.md
```diff
@@ -71,15 +102,24 @@ export JUMPSTARTER_TOKEN=dGhpc2lzYXRva2VuLTEyMzQxMjM0MTIzNDEyMzQtc2Rxd3Jxd2VycXd
...
+run in local mode and create an exporter instance to interact with the hardware.
 
-When client configuration exists
+Communication between the local client and exporter take place over a local
+socket: `/var/run/jumpstarter.sock`.
 
+A local instance of the exporter can also be started using the following command:
 
+```bash
+systemctl start jumpstarter
```
mangelajo, line 124:
```suggestion
systemctl start jumpstarter-exporter
```

just a suggestion :-)

docs/usage.md
kirkbrauer: Yeah, I think standardizing the command like would maybe make sense like: `jumpstarter config set exporter-config /etc/jumpstarter/my-exporter.yaml`

docs/usage.md
kirkbrauer: Added a note in the `architecture.md` about this :)

PR #6, opened 2024-07-16
run pytest on github CI

mangelajo:


.github/workflows/pytest.yaml
```diff
@@ -0,0 +1,37 @@
...
+        uses: abatilo/actions-poetry@v2
+
+      - uses: actions/cache@v3
+        name: Define a cache for the virtual environment based on the dependencies lock file
+        with:
+          path: ./.venv
+          key: venv-${{ hashFiles('poetry.lock') }}
+
+      - name: Fix lock?
+        run: poetry lock
```
mangelajo, line 31: This shouldn't be necessary but for some reason it fails otherwise (may be because of the python version diff?) will need investigation.

PR #8, opened 2024-07-16
Add devcontainer.json configuration

kirkbrauer:
Add VS Code devcontainer configuration files to make local development and testing easier on non-Linux platforms.


mangelajo:
Cool, I haven't actually tried devcontainers, but I will :D

Issue #7, opened 2024-07-16
Define and Implement Jumpstarter CLI Tool : client


kirkbrauer:
Define the usage of and implement a Jumpstarter CLI tool to generate configuration files, execute tests, and manage Jumpstarter service in a k8s cluster.

PR #9, opened 2024-07-17
Add Click CLI for Jumpstarter

kirkbrauer:
Starting work on #7


docs/glossary.md
```diff
@@ -2,33 +2,30 @@
...
 
-## Concepts
+* `Controller`: The central service that authenticates and connects the exporters
+  and clients, manages leases, and provides an inventory of available exporters
+  and clients.
+
+* `Router`: The service that handles routing of messages between clients and
+  exporters once they are authenticated by the controller.
 
-* `Device`: a device that is exposed on a exporter, the exporter enumerates those
```
mangelajo, line 22: In the grpc architecture and objects, the devices are a list under the exporter,
https://github.com/jumpstarter-dev/jumpstarter-protocol/blob/main/proto/jumpstarter/v1/jumpstarter.proto#L120

those are meant to be the different interfaces to the DUT.

I suggest that we split the doc changes to a separate PR so we can discuss them independently from the CLI, the CLI part looks good, I guess we need to figure out the __version__ thing.

kirkbrauer: Ah, I hadn't realized that this change also got bundled into this PR. I'll revert any doc changes here.

docs/usage.md
```diff
@@ -5,7 +5,7 @@
 ### Creating a client token and configuration
 
 ```bash
-jumpstarter create client my-client -o my-client.yaml
+jumpstarter client create my-client -o my-client.yaml
```
mangelajo, line 8: 👍 
Well may be it's good to keep the doc changes related to CLI on the same patch hehe :)

kirkbrauer: Yeah, we can update the `usage.md` to match the intended and actual CLI operation soon.

docs/glossary.md
kirkbrauer: Regarding getting the version. I have seen a few approaches, some tools just use a `VERSION.txt` file that is updated by something external like [`bump2version`](https://pypi.org/project/bump2version/). If we use setuptools as our build tool they also have a way to do this.

mangelajo: one of the advantages of poetry is that we can abstract from setuptools for building, and publishing

I just found out that we can do this now:

>>> import importlib.metadata
>>> importlib.metadata.version("jumpstarter")
'0.1.0'

StackOverflow is always great %) https://stackoverflow.com/questions/74656951/how-do-i-embed-the-version-from-pyproject-toml-so-my-package-can-use-it

kirkbrauer:
Should we move version to it's own subcommand like `kubectl` does? This might make sense as we can ask the server what its version is and print the protocol version as well.

mangelajo:
> Should we move version to it's own subcommand like `kubectl` does? This might make sense as we can ask the server what its version is and print the protocol version as well.

Hmm, that could make sense, eventually we can pull the server version too.

👍

mangelajo (approved):

PR #11, opened 2024-07-25
Add config and driver documentation

kirkbrauer:
- Add documentation about the Jumpstarter configuration files (both user and system level).
- Add documentation on the structure and lifecycle of drivers.


docs/config.md
```diff
@@ -0,0 +1,257 @@
...
+configurations for the exporters.
+
+```yaml
+# /etc/jumpstarter/config.yaml
+
+apiVersion: jumpstarter.dev/v1alpha1
+kind: SystemConfig
+config:
+  current-exporter: ./exporters/myexporter/exporter.yaml
+```
```
mangelajo, line 52: I understood we wanted to remove this, instead we could point the exporters via env var / systemd templates to the wanted config for the exporter instance?

docs/config.md
```diff
@@ -0,0 +1,257 @@
...
+instance, client tokens will be automatically generated using the controller endpoint.
+
+```bash
+$ jmp client create myclient
+Using kubectl context 'my-cluster'
+Generating client auth token
+Created client config '/home/jdoe/.config/jumpstarter/clients/myclient.yaml'
+```
+
+To switch between different client configs, use the `jmp client use <name>` command:
```
mangelajo, line 105: ❤️

docs/config.md
```diff
@@ -0,0 +1,257 @@
...
+
+```bash
+$ jmp exporter start another
+Using exporter config '/etc/jumpstarter/exporters/another/exporter.yaml'
+```
+
+The path to a config can also be provided:
+
+```bash
+$ jmp exporter start --config /etc/jumpstarter/exporters/another/exporter.yaml
```
mangelajo, line 224: We should accept it also as env var

docs/config.md
```diff
@@ -0,0 +1,257 @@
...
+$ jmp exporter use another
+Using exporter config '/etc/jumpstarter/exporters/another/exporter.yaml'
+```
+
+To use a specific config when starting the exporter:
+
+```bash
+$ jmp exporter start another
+Using exporter config '/etc/jumpstarter/exporters/another/exporter.yaml'
+```
```
mangelajo, line 219: I guess if we remove the system exporter ref from L52 those would also go away?

docs/config.md
mangelajo: ah sorry, it's specified down in L249 👍

PR #12, opened 2024-07-26
Add client/user configs via YAML

kirkbrauer:
- Added `config` module with `ClientConfig` and `UserConfig` classes
- Hand-written serialization/deserialization logic
- Created client CLI commands
- Implemented 100% test coverage of configs


mangelajo:
In python 3.11 it doesn't work because typing_extensions is missing, but if we add the package
`poetry add typing_extensions` it works.

Can you add it to your PR? :)

mangelajo (commented):
Looks great, let's see what CI says.

mangelajo (approved):

PR #13, opened 2024-07-29
GPIO Digital{Output,Input} driver

NickCao:


jumpstarter/drivers/gpio/digital.py
```diff
@@ -0,0 +1,62 @@
+from dataclasses import InitVar, dataclass, field
+
+from jumpstarter.drivers import Driver, DriverClient, export
+from gpiozero import DigitalOutputDevice, DigitalInputDevice
```
mangelajo, line 4: gpiozero is oriented to rpi, should we make it gpio/rpi/digital.py  may be?

They we would leave room for other digital io implementations :-)

NickCao: Or shall we group it as `rpi/gpio`, keeping everything rpi specific together?

jumpstarter/drivers/gpio/digital.py
```diff
@@ -0,0 +1,62 @@
...
+    def off(self):
+        self.device.off()
+
+    @export
+    def on(self):
+        self.device.on()
+
+
+@dataclass(kw_only=True)
+class DigitalOutputClient(DriverClient):
```
mangelajo, line 28: nit: it could make sense to separate the client to another module (remove the need to load gpiozero on client side). 

My worry on dependencies comes mostly from security, and load time on smaller systems/exporters.

pyproject.toml
```diff
@@ -15,6 +15,7 @@ pyudev = "^0.24.3"
 httpx = "^0.27.0"
 pyyaml = "^6.0.1"
 anyio = "^4.4.0"
+gpiozero = "^2.0.1"
```
mangelajo, line 18: We will see this in the future, no action required now. But, is there a way to make some dependencies weak, or make specific targets like rpi in this case?

I.e. this one pulls also "colorzero" but we would see more dependency creep in in the future :-)

NickCao: Yes, poetry supports optional dependencies and dependency groups. We can create an optional `rpi` group for everything rpi.

NickCao: Actually, extras not groups: https://python-poetry.org/docs/pyproject/#extras

jumpstarter/drivers/gpio/digital.py
mangelajo: Yeah, that could make more sense from the POV of maintenance too...

jumpstarter/drivers/rpi/gpio/digital_test.py
```diff
@@ -0,0 +1,38 @@
+# ruff: noqa: E402
+from concurrent.futures import ThreadPoolExecutor
+
+import pytest
+
+pytest.importorskip("gpiozero")
```
mangelajo, line 6: Can we install the extras on CI to trigger testing? We can do later, may be we should test with and without the extras to make sure we verify all the possibilities

mangelajo (approved):

jumpstarter/drivers/rpi/gpio/digital_test.py
NickCao: `poetry install --all-extras` would install all these.

PR #17, opened 2024-07-29
Install all optional dependencies in CI

NickCao:


.github/workflows/pytest.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
         run: poetry lock
         
       - name: Install the project dependencies
-        run: poetry install
+        run: poetry install --all-extras
 
       - name: Run the automated tests (for example)
-        run: poetry run pytest -v
\ No newline at end of file
+        run: poetry run pytest -v
```
kirkbrauer, line 37: We can probably eliminate `-v` here with the recent pytest config changes 😄

kirkbrauer (approved):

PR #18, opened 2024-07-29
Set default pytest options to generate coverage report

NickCao:


pyproject.toml
```diff
@@ -60,3 +60,6 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.coverage.run]
 omit = ["conftest.py", "test_*.py", "*_test.py"]
+
+[tool.pytest.ini_options]
+addopts = "--capture=no --verbose --cov --cov-report=html"
```
kirkbrauer, line 65: We should also generate XML: `--cov-report=xml:coverage.xml` for coverage tooks in various IDEs.

NickCao: `--cov-report=xml` is enough, default file name is already `coverage.xml`

kirkbrauer (approved):
Looks great!

PR #21, opened 2024-07-30
Force kw_only for all config classes

kirkbrauer:
Improve config code style to prevent confusion with different arguments.

### DIFF
--- a/jumpstarter/cli/client.py
+++ b/jumpstarter/cli/client.py
@@ -56,7 +56,9 @@ def client_create(
     if out is None and ClientConfig.exists(name):
         raise click.ClickException(f"A client with the name '{name}' already exists.")
 
-    config = ClientConfig(name, endpoint, token, drivers=ClientConfigDrivers(allow.split(","), unsafe))
+    config = ClientConfig(
+        name=name, endpoint=endpoint, token=token, drivers=ClientConfigDrivers(allow=allow.split(","), unsafe=unsafe)
+    )
     ClientConfig.save(config, out)
 
     # If this is the only client config, set it as default
--- a/jumpstarter/config/user.py
+++ b/jumpstarter/config/user.py
@@ -9,7 +9,7 @@ from .client import ClientConfig
 from .common import CONFIG_API_VERSION, CONFIG_PATH
 
 
-@dataclass
+@dataclass(kw_only=True)
 class UserConfig:
     """The user configuration for the Jumpstarter CLI."""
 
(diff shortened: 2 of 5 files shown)


jumpstarter/config/user.py
```diff
@@ -51,14 +51,14 @@ def load() -> Self:
...
-            return UserConfig(current_client)
+            return UserConfig(current_client=current_client)
 
     def load_or_create() -> Self:
         """Check if a user config exists, otherwise create an empty one."""
         if UserConfig.exists() is False:
             if os.path.exists(UserConfig.BASE_CONFIG_PATH) is False:
                 os.makedirs(UserConfig.BASE_CONFIG_PATH)
-            config = UserConfig(None)
+            config = UserConfig(current_client=None)
```
mangelajo, line 61: I would keep it in cases like this, but without enforcing it may be. WDYT?

tests/test_user_config.py
```diff
@@ -65,7 +73,11 @@ def test_user_config_load_current_client_empty():
     current-client: ""
 """
     with patch.object(
-        ClientConfig, "load", return_value=ClientConfig("testclient", "abc", "123", ClientConfigDrivers([], False))
+        ClientConfig,
+        "load",
+        return_value=ClientConfig(
+            name="testclient", endpoint="abc", token="123", drivers=ClientConfigDrivers(allow=[], unsafe=False)
```
mangelajo, line 79: helps here... ':D :-)

jumpstarter/config/user.py
kirkbrauer: Yeah, I'm just worried if we get more properties it'll get confusing.

tests/test_user_config.py
kirkbrauer: I think this makes it look a lot more like the resulting YAML :D

jumpstarter/cli/client.py
```diff
@@ -56,7 +56,9 @@ def client_create(
     if out is None and ClientConfig.exists(name):
         raise click.ClickException(f"A client with the name '{name}' already exists.")
 
-    config = ClientConfig(name, endpoint, token, drivers=ClientConfigDrivers(allow.split(","), unsafe))
+    config = ClientConfig(
+        name=name, endpoint=endpoint, token=token, drivers=ClientConfigDrivers(allow=allow.split(","), unsafe=unsafe)
```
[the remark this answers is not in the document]
kirkbrauer: Good news is that this is internal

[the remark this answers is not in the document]
mangelajo: true :)

mangelajo (approved):
Let's go with it, and let's re-analyze if it got too annoying, as you said, this is internal.

PR #22, opened 2024-07-30
Implement dutlink driver

NickCao:


jumpstarter/drivers/dutlink/base.py
```diff
@@ -0,0 +1,151 @@
...
+from jumpstarter.drivers.storage import StorageMuxInterface
+
+
+@dataclass(kw_only=True)
+class DutlinkPower(Driver):
+    parent: "Dutlink"
+
+    @classmethod
+    def client(cls) -> str:
+        return "jumpstarter.drivers.dutlink.client.DutlinkPowerClient"
```
mangelajo, line 23: Can we provide a generic power control driver client? most power clients won't need anything specific on the client side.

jumpstarter/drivers/dutlink/base.py
```diff
@@ -0,0 +1,151 @@
...
+    async def write(self, src: str):
+        self.control("host")
+
+        with fail_after(20):
+            while True:
+                if os.path.exists(self.storage_device):
+                    try:
+                        Path(self.storage_device).write_bytes(b"\0")
+                    except OSError:
+                        pass  # wait for device ready
```
mangelajo, line 79: we may want to log those events , eventually we should also have a log interface that allows us to send audit events up instead of just console logging :) but we should do that later, not now.

jumpstarter/drivers/dutlink/base.py
```diff
@@ -0,0 +1,151 @@
...
+                    bInterfaceClass=0xFF,
+                    bInterfaceSubClass=0x1,
+                    bInterfaceProtocol=0x1,
+                )
+
+                self.power = DutlinkPower(name="power", parent=self)
+                self.storage = DutlinkStorageMux(name="storage", parent=self, storage_device=self.storage_device)
+
+                udev = pyudev.Context()
+                for tty in udev.list_devices(subsystem="tty", ID_SERIAL_SHORT=serial):
```
mangelajo, line 122: I would treat as error if there is more than one tty device :)

jumpstarter/drivers/dutlink/base.py
```diff
@@ -0,0 +1,151 @@
...
+                    self.console = PySerial(name="console", url=tty.device_node)
+
+                return
+
+        raise FileNotFoundError("failed to find dutlink device")
+
+    @classmethod
+    def client(cls) -> str:
+        return "jumpstarter.drivers.dutlink.client.DutlinkClient"
```
mangelajo, line 131: no need to change it, just trying to understand everything: if we wanted, would it be possible to use jumpstarter.drivers.composite.CompositeClient here without creating a client stub?

jumpstarter/drivers/dutlink/client.py
```diff
@@ -0,0 +1,14 @@
+from jumpstarter.drivers import DriverClient
+from jumpstarter.drivers.composite import CompositeClient
+
+
+class DutlinkPowerClient(DriverClient):
```
mangelajo, line 5: I'd make a generic PowerClient for the power interface :)

jumpstarter/drivers/mixins.py
```diff
@@ -42,6 +47,20 @@ def portforward(self, listener, method="connect"):
             yield
 
 
+class ExpectMixin(StreamMixin):
+    @contextmanager
+    def expect(self):
+        with TemporaryDirectory() as tempdir:
+            socketpath = Path(tempdir) / "socket"
```
mangelajo, line 54: Can you docstring/document  how this works with pyserial, socket, etc... :-)

jumpstarter/drivers/storage/base.py
```diff
@@ -37,8 +37,9 @@ def dut(self):
     def off(self):
         return self.call("off")
 
-    def write(self, src: str):
-        return self.call("write", src)
+    def write(self, filepath):
```
mangelajo, line 40: may be have a write_local and a write for simplicity?

I still think the idea of having different methods to get a file from A to B, or having files from other sources would help in the long term :)

pyproject.toml
```diff
@@ -13,14 +13,20 @@ python = "^3.10"
...
-gpiozero = { version = "^2.0.1", optional = true }
 aiohttp = "^3.9.5"
 
+# rpi
+gpiozero = { version = "^2.0.1", optional = true }
+
+# dutlink
+libusb = {version = "^1.0.27", optional = true}
+pyudev = {version = "^0.24.3", optional = true}
+pexpect = "^4.9.0"
```
mangelajo, line 28:
```suggestion
pexpect = "^4.9.0"
```

This one should probably go along always by default (related to pyserial usage in general)

jumpstarter/drivers/dutlink/base.py
NickCao: Somehow I can no longer reproduce this race condition! But yesterday there was always a chance of the device file being present but can't be written to.

jumpstarter/drivers/dutlink/base.py
NickCao: Yes, actually let's change this so dutlink no longer needs any custom client stub.

pyproject.toml
NickCao: `poetry add` always adds things to the botton...

jumpstarter/drivers/dutlink/base_test.py
```diff
@@ -39,5 +26,22 @@ def test_drivers_dutlink():
                     s.connect(str(socketpath))
 
                     expect = fdspawn(s)
+                    expect.send("\x02" * 5)
+
                     expect.send("about\r\n")
                     expect.expect("Jumpstarter test-harness")
+
+                    expect.send("console\r\n")
```
mangelajo: we should probably ditch the console mode on jumpstarter, ... make it always be the connection to the dut, and have everything be handled via USB control commands.

jumpstarter/drivers/dutlink/base.py
```diff
@@ -0,0 +1,155 @@
...
+
+from jumpstarter.drivers import Driver, export
+from jumpstarter.drivers.composite import CompositeInterface
+from jumpstarter.drivers.power import PowerInterface, PowerReading
+from jumpstarter.drivers.serial.pyserial import PySerial
+from jumpstarter.drivers.storage import StorageMuxInterface
+
+
+@dataclass(kw_only=True)
+class DutlinkPower(PowerInterface, Driver):
```
mangelajo, line 21: 👍

jumpstarter/drivers/dutlink/base_test.py
NickCao: The console mode is actually helpful here, otherwise we'd require a connected DUT to actually test the console connection.

NickCao:
Ready for another round of review. Also fixed the `j` magic command.

jumpstarter/cli/shell.py
```diff
@@ -19,13 +16,9 @@ async def shell_impl():
 
     session = Session(
         name="transient",
-        root_device=Composite(
```
mangelajo, line 22: Why do we change it here to Dutlink?

jumpstarter/cli/control.py
```diff
@@ -1,28 +0,0 @@
-import os
```
mangelajo, line 1: Why do we remove the control command?

mangelajo (commented):
A couple of questions, other than that it looks good.

jumpstarter/cli/control.py
NickCao: It's move to the new `j` command. Since this part actually required async unlike other parts of the cli.

jumpstarter/cli/shell.py
NickCao: Oh that was for testing.

NickCao: Reverted

mangelajo (approved):

PR #29, opened 2024-08-01
Explicitly signal stream initiation and termination

NickCao:


jumpstarter/drivers/base.py
```diff
@@ -98,12 +98,14 @@ async def Stream(self, request_iterator, context):
...
                 await resource.send(str(resource_uuid).encode("utf-8"))
+                await resource.send_eof()
 
                 async with remote:
                     async for v in forward_server_stream(request_iterator, remote):
                         yield v
 
-                del self.resources[resource_uuid]
+                # del self.resources[resource_uuid]
+                # small resources might be fully buffered in memory
```
mangelajo, line 108: What do you mean by this?

jumpstarter/common/streams.py
```diff
@@ -1,58 +1,70 @@
...
+                case router_pb2.FRAME_TYPE_DATA:
+                    await tx.send(frame.payload)
+                case router_pb2.FRAME_TYPE_GOAWAY:
+                    if isinstance(tx, ObjectStream) or isinstance(tx, ByteStream):
+                        await tx.send_eof()
+                case _:
+                    pass
+    # ignore peer disconnect
+    except BrokenResourceError:
+        pass
```
mangelajo, line 34: we are doing excessive use of pass , in some cases we should at least log it (warning), which will help in the future if issues arise from it.

jumpstarter/common/streams.py
```diff
@@ -1,58 +1,70 @@
...
+    # ignore rpc cancellation and internal error
+    except grpc.aio.AioRpcError as e:
+        match e.code():
+            case grpc.StatusCode.CANCELLED | grpc.StatusCode.INTERNAL:
                 pass
```
mangelajo, line 39: same here.

jumpstarter/common/streams.py
```diff
@@ -1,58 +1,70 @@
...
+    try:
+        async for frame in rx:
+            match frame.frame_type:
+                case router_pb2.FRAME_TYPE_DATA:
+                    await tx.send(frame.payload)
+                case router_pb2.FRAME_TYPE_GOAWAY:
+                    if isinstance(tx, ObjectStream) or isinstance(tx, ByteStream):
+                        await tx.send_eof()
+                case _:
+                    pass
```
mangelajo, line 31: see my comments on pass.

mangelajo (changes requested):
Looks great, just some comments about logging all "pass calls"

jumpstarter/drivers/base.py
NickCao: Sequence of event:
1. client opens a stream for sharing resource
2. exporter accepts the stream and registers the other end of the stream and a resource
3. exporter starts task to forward data from the client side of the stream to the exporter side of the stream
3.1 (actually happened) stream is copied into internal buffer before being consumed, forward tasks exits, resource unregistered
4. driver get the stream from the dict and begins consuming the stream
4.1 (actually happened) driver fails to find resource
6. (originally intended) stream is fully consumed by the driver, forward tasks exits, resource unregistered

NickCao: TLDR: resource shouldn't be unregistered when forwarding task exits, but only when it's fully consumed (or client decides to cancel)

jumpstarter/common/streams.py
NickCao: Opted for debug level, warning is too strong for these (expected to happen in daily use) errors.

jumpstarter/drivers/base.py
mangelajo: Thanks for the explanation I see now :-)

mangelajo (approved):

PR #42, opened 2024-08-08
Client config test rework

NickCao:


tests/test_client_config.py
```diff
@@ -70,17 +61,13 @@ def test_client_config_from_env_allow_unsafe():
     assert config.drivers.unsafe is True
 
 
-def test_client_config_from_env_no_token_raises():
-    os.environ[JMP_ENDPOINT] = "grpcs://jumpstarter.my-lab.com:1443"
-    os.environ[JMP_DRIVERS_ALLOW] = "jumpstarter.drivers.*,vendorpackage.*"
+@pytest.mark.parametrize("missing_field", [JMP_TOKEN, JMP_ENDPOINT])
+def test_client_config_from_env_missing_field_raises(monkeypatch, missing_field):
+    monkeypatch.setenv(JMP_TOKEN, "dGhpc2lzYXRva2VuLTEyMzQxMjM0MTIzNEyMzQtc2Rxd3Jxd2VycXdlcnF3ZXJxd2VyLTEyMzQxMjM0MTIz")
```
mangelajo, line 66: nice I didn't know we had a test patcher for env :D

tests/test_client_config.py
```diff
@@ -109,122 +96,80 @@ def test_client_config_from_file():
...
+        "client": {
+            "endpoint": "grpcs://jumpstarter.my-lab.com:1443",
+            "token": "dGhpc2lzYXRva2VuLTEyMzQxMjM0MTIzNEyMzQtc2Rxd3Jxd2VycXdlcnF3ZXJxd2VyLTEyMzQxMjM0MTIz",
+            "drivers": {"allow": ["jumpstarter.drivers.*", "vendorpackage.*"]},
+        },
+    }
+
+    del CLIENT_CONFIG[missing_field]
+    with tempfile.NamedTemporaryFile(mode="w") as f:
+        yaml.safe_dump(CLIENT_CONFIG, f, sort_keys=False)
```
mangelajo, line 132: I was chiming here because we were stopping to test/enforce the in-disk format for yaml, but I see we have checkers for that around L199

tests/test_client_config.py
```diff
@@ -109,122 +96,80 @@ def test_client_config_from_file():
...
+            "drivers": {"allow": ["jumpstarter.drivers.*", "vendorpackage.*"]},
+        },
+    }
+
+    del CLIENT_CONFIG["client"][missing_field]
+    with tempfile.NamedTemporaryFile(mode="w") as f:
+        yaml.safe_dump(CLIENT_CONFIG, f, sort_keys=False)
         with pytest.raises(ValueError):
             _ = ClientConfig.from_file(f.name)
-        os.unlink(f.name)
```
mangelajo, line 159: Are we sure it's deleted? I added this because I was seeing the files not being deleted in my env.

NickCao: It's indeed cleaned up, I find out the offending test is `test_client_ensure_exists_makes_dir`

tests/test_client_config.py
NickCao: We are still testing that as part of `test_client_config_from_file`

tests/test_client_config.py
NickCao: Ah the problem is even deeper, since CLIENT_CONFIGS_PATH is a class attribute, the tempdir path is being leaked to other tests.

kirkbrauer:
Yeah, I thought some of these tests might be leaky. We might want to refactor this further to avoid further problems with the tests.

mangelajo (approved):
Nice 👍🏻

PR #43, opened 2024-08-09
Make server stream a context manager, implement keepalive

NickCao:
The continuation of explicit stream cancellation work, now server streams (driver side of the stream) can also be cancelled. This also fully mitigates the pitfall described in https://anyio.readthedocs.io/en/stable/cancellation.html#avoiding-cancel-scope-stack-corruption

### DIFF
--- a/jumpstarter/common/streams.py
+++ b/jumpstarter/common/streams.py
@@ -2,8 +2,8 @@ import logging
 from contextlib import asynccontextmanager
 
 import grpc
-from anyio import BrokenResourceError, ClosedResourceError, create_memory_object_stream, create_task_group
-from anyio.abc import ByteStream, ObjectStream
+from anyio import BrokenResourceError, EndOfStream, create_memory_object_stream, create_task_group
+from anyio.abc import AnyByteStream, ByteStream, ObjectStream
 from anyio.streams.stapled import StapledObjectStream
 
 from jumpstarter.v1 import router_pb2, router_pb2_grpc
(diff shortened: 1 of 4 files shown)


jumpstarter/common/streams.py
```diff
@@ -1,68 +1,109 @@
...
+    get_cancelled_exc_class,
+    move_on_after,
+)
+from anyio.abc import AnyByteStream, ByteStream, ObjectStream
 from anyio.streams.stapled import StapledObjectStream
 
 from jumpstarter.v1 import router_pb2, router_pb2_grpc
 
-logger = logging.getLogger(__name__)
+KEEPALIVE_INTERVAL = 1
```
mangelajo, line 20: Should we make this a bit more tolerant? 5s and 30s maybe?

just a gut feeling

NickCao: We can make it `KEEPALIVE_INTERVAL = 1 hour`, `KEEPALIVE_TOLERANCE= another 1 hour`?

NickCao: Overall this is more about our desired use case: whether the keepalive is for reaping long gone connections, or to ensure the peer is responsive in a timely manner.

mangelajo: I’d make it in minutes and let control through env , then we can experiment

jumpstarter/common/streams.py
```diff
@@ -1,68 +1,110 @@
...
+    get_cancelled_exc_class,
+    move_on_after,
+)
+from anyio.abc import AnyByteStream, ByteStream, ObjectStream
 from anyio.streams.stapled import StapledObjectStream
 
 from jumpstarter.v1 import router_pb2, router_pb2_grpc
 
-logger = logging.getLogger(__name__)
+KEEPALIVE_INTERVAL = int(os.environ.get("JMP_KEEPALIVE_INTERVAL", "5")) * 60
```
mangelajo, line 21: Why do we force it to be minutes?, I would leave the freedom for it to be seconds too. May be the available hardware is short and you want connections to be released asap?.

mangelajo (approved):

PR #44, opened 2024-08-12
Enforce the use of pydantic over grpc

NickCao:
Resend of #38 targeting `main`, that one was targeting `opendal` due to dependency on #36


mangelajo:
> Resend of #38 targeting `main`, that one was targeting `opendal` due to dependency on #36

oops :D

mangelajo (approved):

PR #45, opened 2024-08-12
Reimplement ClientConfig with pydantic

NickCao:
We also need to think about future config version updates.

Proposal:
Support loading old versions, only support writing latest version. When saving old versions, transparently upgrade them to the latest version.

### DIFF
--- a/jumpstarter/cli/client.py
+++ b/jumpstarter/cli/client.py
@@ -2,7 +2,7 @@ from typing import Optional
 
 import click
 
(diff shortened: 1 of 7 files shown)


kirkbrauer:
@NickCao I agree with the versioning, however if we add optional fields, I don't think we need to create a new version of the config, just implement the new defaults.

mangelajo (approved):

mangelajo (approved):

PR #47, opened 2024-08-13
Use pydantic in ustreamer driver

NickCao:


jumpstarter/drivers/video/common.py
```diff
@@ -0,0 +1,33 @@
+from pydantic import BaseModel
+
+
+class UStreamerState(BaseModel):
```
mangelajo, line 4: nit: why common if it's UStreamer specific?

mangelajo (approved):

mangelajo:
auto merging, my comment was a nit..

PR #51, opened 2024-08-21
Switch from `poetry` to `uv` package manager

kirkbrauer:
Proposal to switch from poetry to [uv](https://docs.astral.sh/uv/) as our package manager.

uv has the following benefits:
- Support for cargo-style [workspaces](https://docs.astral.sh/uv/concepts/workspaces/) that enable us to have a monorepo.
- Follows Python standards wherever possible.
- Installs and [manages Python versions](https://docs.astral.sh/uv/#python-management)
- Provides a [cross-platform lockfile](https://docs.astral.sh/uv/concepts/projects/#lockfile).
- Much faster installs (they claim [10-100x faster](https://github.com/astral-sh/uv/blob/main/BENCHMARKS.md)).


kirkbrauer:
Closing in favor of #52

PR #53, opened 2024-08-21
Reorganize client and driver base classes to their own packages

NickCao:


mangelajo:
It looks good, I like the reorg :)

PR #56, opened 2024-08-22
Make `children` a field of `Driver` and `DriverClient` base class

NickCao:
This allows us to automatically set the `jumpstarter.dev/name` well-known label, and simplifies the implementation of composite harness.

Depends on https://github.com/jumpstarter-dev/jumpstarter/pull/53


mangelajo (approved):
looks good

PR #52, opened 2024-08-21
Make use of uv workspace

NickCao:

PR #57, opened 2024-08-22
Lazily wrap driver method

NickCao:
This enables drivers to be used directly without going through grpc. Eases testing driver directly.

Depends on https://github.com/jumpstarter-dev/jumpstarter/pull/56


NickCao:
Merged as part of #52

PR #58, opened 2024-08-26
Initial can driver

NickCao:


kirkbrauer (commented):
Looks good so far, I'll do some additional testing with CAN

NickCao:
> Looks good so far, I'll do some additional testing with CAN

We can parametrize the Driver initialization part to make use of real can interfaces.

PR #70, opened 2024-09-05
Add sync target

NickCao:


kirkbrauer:
Nice! I think this is a much better way to solve this.

kirkbrauer (approved):

PR #72, opened 2024-09-05
Implement Isotp driver

NickCao:


mangelajo:
@kirkbrauer I leave it for you to review/merge, I am unfamiliar with this.

kirkbrauer:
@mangelajo Sure, the gist is that this is a protocol similar to TCP that sits on top of CAN as the base transport. It just allows us to send large "packets" to data such as an image that we flash. Originally, Nick had implemented it as a client-side driver, but the issue is the latency could mess with the stringent timing requirements. The solution was to send the entire payload via gRPC and then rely on the Linux kernel driver (or user space fallback) on the exporter side to keep the timing.

contrib/can/jumpstarter_driver_can/driver.py
```diff
@@ -94,3 +95,150 @@ def flush_tx_buffer(self) -> None:
     @validate_call(validate_return=True)
     def shutdown(self) -> None:
         self.bus.shutdown()
+
+
+@dataclass(kw_only=True, config=ConfigDict(arbitrary_types_allowed=True))
+class Isotp(Driver):
```
kirkbrauer, line 101: We should probably use the capitalization `IsoTp` for consistency.

contrib/can/jumpstarter_driver_can/driver.py
```diff
@@ -94,3 +95,150 @@ def flush_tx_buffer(self) -> None:
...
+        self.stack.stop_sending()
+
+    @export
+    @validate_call(validate_return=True)
+    def stop_receiving(self) -> None:
+        self.stack.stop_receiving()
+
+
+@dataclass(kw_only=True, config=ConfigDict(arbitrary_types_allowed=True))
+class IsotpSocket(Driver):
```
kirkbrauer, line 177: We might want to think about better naming to ensure that it is clear what each kind of ISO-TP implementation means for performance and compatibility. It probably wouldn't hurt to maybe add some documentation comments.

kirkbrauer (approved):

PR #74, opened 2024-09-12
Implement safe import_class

NickCao:


jumpstarter/common/importlib.py
```diff
@@ -0,0 +1,30 @@
+# Reference: https://docs.djangoproject.com/en/5.0/_modules/django/utils/module_loading/#import_string
+
+import sys
+from fnmatch import fnmatchcase
+from importlib import import_module
+
+
+def cached_import(module_path, class_name):
```
mangelajo, line 8: Could you extend and explain in docstring how it works and how do we use it?

NickCao: Added two examples.

mangelajo (approved):
Thank you!,

PR #80, opened 2024-09-13
Add examples directory, targets, and example

kirkbrauer:
To faciliate the easier creation of examples in the future, I have added an `examples` directory and configs in the `Makefile` and `pyproject.toml` to faciliate this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,14 +39,7 @@ build-contrib: $(addprefix build-,$(CONTRIB_TARGETS))
 sync-example-%: examples/%
 	uv sync --all-extras --inexact --package jumpstarter_example_$(<F)
 
-test-example-%: examples/%
-	uv run --isolated --package jumpstarter_example_$(<F) pytest $<
-
 sync-examples: $(addprefix sync-,$(EXAMPLE_TARGETS))
-
-test-examples: $(addprefix test-,$(EXAMPLE_TARGETS))
-
-build-examples: $(addprefix build-,$(EXAMPLE_TARGETS))
 
 clean-venv:
 	-rm -rf ./.venv
@@ -62,10 +55,10 @@ clean-test:
 
 sync: sync-jumpstarter sync-contrib sync-examples
 
-test: test-jumpstarter test-contrib test-examples
+test: test-jumpstarter test-contrib
 
 build: sync build-jumpstarter build-contrib
 
 clean: clean-docs clean-venv clean-build clean-test
 
-.PHONY: sync docs test test-jumpstarter test-contrib build build-jumpstarter build-contrib test-examples clean-test clean-docs clean-venv clean-build
+.PHONY: sync docs test test-jumpstarter test-contrib build build-jumpstarter build-contrib clean-test clean-docs clean-venv clean-build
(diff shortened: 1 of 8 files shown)


kirkbrauer:
Let's try to merge this so we have a template for adding future examples to the project.

examples/automotive/pyproject.toml
```diff
@@ -0,0 +1,21 @@
...
+dependencies = [
+    "jumpstarter"
+]
+
+[tool.uv.sources]
+jumpstarter = { workspace = true }
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
```
NickCao, line 21: We can omit the `build-system` section since this is a loosely organized application instead of a "packaged" application: https://docs.astral.sh/uv/concepts/projects/#applications

Makefile
```diff
@@ -35,6 +36,21 @@ test-contrib: $(addprefix test-,$(CONTRIB_TARGETS))
...
 build-contrib: $(addprefix build-,$(CONTRIB_TARGETS))
 
+sync-example-%: examples/%
+	uv sync --all-extras --inexact --package jumpstarter_example_$(<F)
+
+test-example-%: examples/%
+	uv run --isolated --package jumpstarter_example_$(<F) pytest $<
+
+build-example-%: examples/%
+	uvx --from build pyproject-build --installer uv --outdir dist $<
```
NickCao, line 46: Same here, we are not building wheels for examples.

NickCao:
We are seeing test failures on contrib packages.
```
error: Failed to spawn: `pytest`
  Caused by: No such file or directory (os error 2)
```
Likely due to https://github.com/astral-sh/uv/releases/tag/0.4.10

NickCao:
> We are seeing test failures on contrib packages.
> 
> ```
> error: Failed to spawn: `pytest`
>   Caused by: No such file or directory (os error 2)
> ```
> 
> Likely due to https://github.com/astral-sh/uv/releases/tag/0.4.10

Namely: Avoid installing transitive dev dependencies

NickCao:
See https://github.com/jumpstarter-dev/jumpstarter/pull/81

NickCao:
The same applies to the new example package. Do we actually need to run pytest on that?

kirkbrauer:
Humm, still seems to be failing to discover the tests

kirkbrauer:
@NickCao we can probably skip for now as it will probably consist of tests haha. We probably don't want to actually run those when we're not connected to the DUT.

Makefile
```diff
@@ -35,6 +36,18 @@ test-contrib: $(addprefix test-,$(CONTRIB_TARGETS))
 
 build-contrib: $(addprefix build-,$(CONTRIB_TARGETS))
 
+sync-example-%: examples/%
+	uv sync --all-extras --inexact --package jumpstarter_example_$(<F)
+
+test-example-%: examples/%
+	uv run --isolated --package jumpstarter_example_$(<F) pytest $<
```
NickCao, line 43: Then just drop the test-example targets.

NickCao (approved):

PR #91, opened 2024-09-24
Initial sd-wire driver

NickCao:


contrib/sdwire/pyproject.toml
```diff
@@ -0,0 +1,28 @@
+[project]
+name = "jumpstarter-driver-sdwire"
+version = "0.1.0"
+description = ""
+authors = [
+  { name = "Miguel Angel Ajo Pelayo", email = "majopela@redhat.com" },
+  { name = "Nick Cao", email = "ncao@redhat.com" },
+  { name = "Kirk Brauer", email = "kbrauer@hatci.com" },
```
mangelajo, line 8:
```suggestion
  { name = "Nick Cao", email = "ncao@redhat.com" },
```

Let's be honest!, good work!

mangelajo:
Nice, so you "hacked" the control messages already? %)

NickCao:
> Nice, so you "hacked" the control messages already? %)

The official cli is complicated since it has to support three different devices, sd-wire/sd-mux/usb-mux, but we are only targeting the simplest of these three :)

contrib/sdwire/jumpstarter_driver_sdwire/driver.py
```diff
@@ -0,0 +1,97 @@
...
+        self.host()
+
+    @export
+    async def write(self, src: str):
+        self.host()
+
+        with fail_after(10):
+            while True:
+                # https://stackoverflow.com/a/2774125
+                fd = os.open(self.storage_device, os.O_WRONLY)
```
mangelajo, line 86: The storage device name can change over time

https://github.com/redhat-et/jumpstarter/blob/main/pkg/drivers/sd-wire/udev.go#L58

This is how we did it on the go version of the drivers, not sure if completely failproof, but it worked.

The problem is that linux will rotate the disk device (/dev/sd{a,b,c,d,..} over time sometimes (not very deterministic).

mangelajo (changes requested):
It looks good, just a small comment related to sdwire from the issues we found on the golang implementation.

We may need to look at that to have something stable.

contrib/sdwire/jumpstarter_driver_sdwire/driver.py
mangelajo: May be we should auto-detect using that method when nothing has been provided, but still leave the storage_device parameter for other systems (i.e. it could potentially work on mac OS)

NickCao: > The storage device name can change over time

We can always use stable device names under ` /dev/disk/by-id/`

mangelajo: that doesn't work because sd-wire does not seem to expose the disk with a id different among instances of sdwire (I tried with 2 sdwires in 1 host)

NickCao: Implemented auto discovery via udev hierarchy, also uses `/dev/disk/by-diskseq` to ensure uniqueness.

mangelajo (approved):
looks great! :)

PR #96, opened 2024-09-26
Generate python package index as part of doc

NickCao:


NickCao:
With this you can install packages as such
```shell
pip install --extra-index-url https://docs.jumpstarter.dev/packages/simple jumpstarter-driver-can
```

mangelajo (approved):
I think this is great until we have time to submit to pypi or setup a custom subdomain.

PR #120, opened 2024-10-09
soc testing example usign pytest and imagehash

mangelajo:
Provides a SOC testing example involving drivers like dutlink and ustreamer, console, storage, etc..


contrib/dutlink/examples/rpi4_video.py
```diff
@@ -0,0 +1,104 @@
...
+    with PexpectAdapter(client=dutlink.console) as console:
+        # stream console output to stdout
+        console.logfile = sys.stdout.buffer
+        # ensure DUT is powered off
+        dutlink.power.off()
+        with open("camera_off.jpeg", "wb") as f:
+            f.write(camera.snapshot())			
+        click.secho("Writing system image", fg="red")
+
+#        dutlink.storage.write_local_file("2024-07-04-raspios-bookworm-arm64-lite.img")
```
mangelajo, line 29:
```suggestion
        dutlink.storage.write_local_file("2024-07-04-raspios-bookworm-arm64-lite.img")
```

I should probably include the script to download/prepare this image.

contrib/ustreamer/jumpstarter_driver_ustreamer/client.py
```diff
@@ -14,4 +25,64 @@ def state(self):
...
     def snapshot(self):
+        """
+        Get a snapshot image from the video input
+
+        :return: JPEG image data
+        :rtype: bytes
+        """
         return b64decode(self.call("snapshot"))
+
+    def write_snapshot(self, filename):
```
mangelajo, line 36:
```suggestion
    def write_snapshot(self, filename):
```

I wonder if we should just have a separate module

jumpstarter_imagehash

Providing an adapter to video Inputs:

vhash = ImageHashAdapter(client.video)

vhash.snapshot()
vhash.hash_snapshot()
vhash.assert_snapshot(....)


In this way we'd keep this functionality shared to other potential video drivers, 
and we would keep away imagehash dependency (+scipy, etc..) from ustreamer.

contrib/dutlink/examples/pytest/test_on_rpi4.py
```diff
@@ -0,0 +1,110 @@
...
+            cls._client.dutlink.power.off()
+        except Exception as e:
+            log.error(f"Failed to power off the device: {e}")
+
+        # call parent teardown
+        super().teardown_class()
+
+    @pytest.fixture()
+    def client(self):
+        return self._client
```
mangelajo, line 29:
```suggestion
```
This one can be removed, it's provided by JumpstarterTest base class..

contrib/ustreamer/jumpstarter_driver_ustreamer/client.py
```diff
@@ -14,4 +25,66 @@ def state(self):
         return UStreamerState.model_validate(self.call("state"))
 
     def snapshot(self):
+        """
+        Get a snapshot image from the video input
+
+        :return: JPEG image data
+        :rtype: bytes
+        """
         return b64decode(self.call("snapshot"))
```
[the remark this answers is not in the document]
mangelajo: Hmm, that could make sense, yes.

contrib/ustreamer/jumpstarter_driver_ustreamer/client.py
mangelajo: Well, this adaptor would require snapshot, we can assume that the video stream is optional :)

contrib/ustreamer/jumpstarter_driver_ustreamer/client.py
```diff
@@ -14,4 +25,66 @@ def state(self):
...
+
+    def write_snapshot(self, filename):
+        """
+        Write a snapshot image into a disk file
+
+        :param str filename: file name to write the snapshot image
+        """
+        with open(filename, "wb") as f:
+            f.write(self.snapshot())
```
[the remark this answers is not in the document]
mangelajo: 👍

jumpstarter/testing/pytest.py
```diff
@@ -0,0 +1,63 @@
...
+            client.dutlink.power.off()
+            log.info("Setting up device")
+            client.dutlink.storage.write_local_file("2024-07-04-raspios-bookworm-arm64-lite.img")
+            client.dutlink.storage.dut()
+            client.dutlink.power.on()
+    ```
+    """
+    @classmethod
+    def setup_class(cls):
+        if os.getenv("JUMPSTARTER_HOST"):
```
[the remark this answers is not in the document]
mangelajo: yes, I had that at the start but catching will be empty (not doing the client)

otherwise if the config file load fails.. the exception is horrible:

"Exception was thrown (big dump) while recovering from this other Exception.... (big stack dump)"

pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+addopts = -s --ignore contrib/dutlink/examples/pytest/test_on_rpi4.py
```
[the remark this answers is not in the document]
mangelajo: I mean it fails if you run it to through the makefile too, when you try to run the individual sub-package pytest, it will look for this parent setting, I will try by embedding it onto the pyproject of the subproject; may be that works, but as pytest.ini it didn't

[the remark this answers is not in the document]
mangelajo: ahh it worked on the pyproject.yaml

examples/soc-pytest/pyproject.toml
```diff
@@ -0,0 +1,28 @@
...
+  { name = "Nick Cao", email = "ncao@redhat.com" },
+]
+readme = "README.md"
+requires-python = ">=3.11"
+dependencies = [
+    "jumpstarter",
+    "pytest>=8.3.2",
+    # TODO: figure this out
+    #"jumsptarter_imagehash",
+    #"jumpstarter_driver_dutlink",
```
NickCao, line 17: What's not working?

jumpstarter/testing/utils.py
```diff
@@ -0,0 +1,22 @@
...
+    Wait for login prompt and login
+
+    :param pexpect_console: pexpect console object
+    :type pexpect_console: pexpect.spawn
+
+    :return: pexpect console object
+    :rtype: pexpect.spawn
+    """
+    log.info("Waiting for login prompt")
+    pexpect_console.expect("login:", timeout=timeout)
```
NickCao, line 16: This should be part of the example, the login prompts could be different across distributions.

kirkbrauer (approved):

examples/soc-pytest/jumpstarter_example_soc_pytest/test_on_rpi4.py
```diff
@@ -0,0 +1,111 @@
...
+    def test_power_on_camera(self, client):
+        client.dutlink.power.on()
+        time.sleep(1)
+        client.camera.snapshot().save("camera_on.jpeg")
+        client.dutlink.power.off()
+
+    def test_power_on_hdmi(self, client, video):
+        # check all the image snapshots through the rpi4 boot process
+        client.dutlink.power.on()
+        time.sleep(1)
```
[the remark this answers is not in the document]
mangelajo: surprisingly reliable, but yes, this should not be the end goal :)

just a starting point. As we said, something that is going to compare video input and expect images to appear in a specific order, and finally provide a list of appearances, and timestamps, probably running on the exporter itself?

examples/soc-pytest/pyproject.toml
mangelajo: If you can try to uncomment it locally, it doesn't seem to work.

jumpstarter/testing/utils.py
mangelajo: I am OK with moving it to the example, but we should have a generic shell driver/adapter in the long term to help us drive the login/shell interaction with a common interface. 

Probably to put in in the core we should make it ready to handle all those other case

jumpstarter/testing/utils.py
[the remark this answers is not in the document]
mangelajo: I'll move it into the example itself, until we can create a more common adapter for driving shell/login interaction.

mangelajo:
I'll merge for now and follow up with your comments in a separate PR, thanks for reviewing and the meaningful requests.

examples/soc-pytest/pyproject.toml
NickCao: It was a typo: https://github.com/jumpstarter-dev/jumpstarter/pull/132

PR #123, opened 2024-10-10
Make use of new status method

NickCao:


jumpstarter/exporter/exporter.py
```diff
@@ -65,6 +49,12 @@ async def handle(self):
             labels=self.labels,
             root_device=self.device_factory(),
         ) as session:
+            await self.Register(
```
mangelajo: Can we keep the logger.info("Registering exporter with controller") here?

jumpstarter/client/lease.py
```diff
@@ -24,17 +24,18 @@ class LeaseRequest(AbstractContextManager, AbstractAsyncContextManager):
     channel: Channel
     metadata_filter: MetadataFilter
     portal: BlockingPortal
+    stub: jumpstarter_pb2_grpc.ControllerServiceStub = field(init=False)
 
     def __post_init__(self):
-        jumpstarter_pb2_grpc.ControllerServiceStub.__init__(self, self.channel)
+        self.stub = jumpstarter_pb2_grpc.ControllerServiceStub(self.channel)
```
mangelajo, line 29: may be we should call those self.controller ? 

Then it's more obvious for readers later in calls self.controller.RequestLease , self.controller.$method,

jumpstarter/exporter/session.py
```diff
@@ -43,18 +43,22 @@ def __init__(self, *args, root_device, **kwargs):
         self.mapping = {u: i for (u, _, _, i) in self.root_device.enumerate()}
 
     @asynccontextmanager
-    async def serve_unix_async(self):
+    async def serve_port_async(self, port):
```
mangelajo, line 46: what are you plotting? :-D :-)  it's good though.

mangelajo (approved):
looks good just some very nit commits, feel free to handle or not, just if you think it they make sense, none is really very important.

mangelajo (approved):

NickCao:
Got some race condition on the controller side: https://github.com/jumpstarter-dev/jumpstarter-controller/actions/runs/11294698671/job/31416030560#step:9:75

The client could be too eager to connect to a newly acquired lease that the exporter is not even listening yet.

NickCao:
> Got some race condition on the controller side: https://github.com/jumpstarter-dev/jumpstarter-controller/actions/runs/11294698671/job/31416030560#step:9:75
> 
> The client could be too eager to connect to a newly acquired lease that the exporter is not even listening yet.

Resolved by allowing the client to wait for the exporter: https://github.com/jumpstarter-dev/jumpstarter-controller/pull/53

PR #129, opened 2024-10-16
Create the jumpstarter_imagehash library

mangelajo:
Provides an aditional ImageHash adapter that builds on top of video/snapshot interfaces providing the following additional methods:

 * hash_snapshot
 * assert_snapshot

uses the imagehash library dependency for simple
asserting/hashing/comparisons of images.

This patch also moves existing contrib drivers into contrib/drivers and leaves room for contrib/libs so we can use that for anything non-driver.


contrib/drivers/ustreamer/jumpstarter_driver_ustreamer/client.py
```diff
@@ -0,0 +1,34 @@
+import io
+import logging
+from base64 import b64decode
+
+from PIL import Image
+
+from jumpstarter.client import DriverClient
```
mangelajo, line 7: this is only adding docstrings, for some reason git didn't detect my move+edit, and took it as delete/create

contrib/libs/imagehash/jumpstarter_imagehash/imagehash.py
```diff
@@ -0,0 +1,71 @@
+import logging
+import os
+
+import imagehash
+from PIL import Image
+
+log = logging.getLogger("imagehash")
+
+class ImageHash:
```
NickCao, line 9: This could be a dataclass.

contrib/libs/imagehash/jumpstarter_imagehash/imagehash.py
```diff
@@ -0,0 +1,71 @@
...
+        Assert the snapshot image is the same as the reference image
+
+        :param str reference_img_file: reference image file name
+        :param int tolerance: hash difference tolerance
+
+        :raises AssertionError: if the snapshot image is different from the reference image
+        """
+        diff, snapshot_img  = self._snapshot_diff(reference_img_file)
+        if diff > tolerance:
+            directory = os.getcwd()
```
NickCao, line 55: The directory should be configurable (default to cwd)

NickCao (approved):
LGTM.

PR #158, opened 2024-10-24
Add guide for writing drivers

NickCao:


kirkbrauer:
@NickCao I made some improvements to the wording and make sure that the intro is more consistent with these new docs.

docs/source/getting-started/custom-driver.md
```diff
@@ -0,0 +1,127 @@
...
+
+```{note}
+It is important that your driver client/exporter package versions are compatible
+or backwards-compatible. We recommend treating drivers APIs like Protobufs and
+versioning in a backwards-compatible way to prevent issues.
+```
+
+## Using a Driver
+
+Custom drivers can be used in the same way as Jumpstarter's built-in driver packages.
```
mangelajo, line 107: We may want to extend this with an example on how to add the driver to the existing exporter image. (follow up)

docs/source/getting-started/custom-driver.md
```diff
@@ -0,0 +1,127 @@
...
+token: ""
+export:
+    custom:
+        # Full import path of the driver class
+        type: jumpstarter_driver_custom.CustomDriver
+```
+
+## Advanced Drivers
+
+Drivers can also take configuration parameters, export generator methods,
```
mangelajo, line 125: I think we may also want to include some example of drivers that take arguments/generators. (follow up)

docs/source/reference/drivers.md
```diff
@@ -1,3 +1,4 @@
+(driver-api)=
```
mangelajo, line 1: what does this mean?

mangelajo (approved):
Nice work, some comments for improvement/follow up. Merging

docs/source/reference/drivers.md
NickCao: It's an explicit cross reference target: https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#creating-explicit-targets

mangelajo: ahh thanks :)

PR #161, opened 2024-10-29
Devspaces support

mangelajo:


Makefile
```diff
@@ -1,17 +1,15 @@
...
 default: build
 
 docs: sync
 	cd docs && uv run make html
 
 serve-docs: sync
-	cd docs && uv run make serve
-
-watch-docs:
-	sphinx-autobuild docs/source docs/build/html
```
mangelajo, line 14: serve-docs was already doing the sphinx-autobuild

NickCao:
Is it possible to use devspace in github codespace? So we don't have to maintain two sets of dev environments.

mangelajo:
> Is it possible to use devspace in github codespace? So we don't have to maintain two sets of dev environments.

Not yet, I have hope that all this will converge over time.

.devcontainer/Dockerfile
```diff
@@ -37,3 +37,6 @@ RUN chown -R vscode $CARGO_HOME
 
 # Install required tools for development
 RUN apt-get update && apt-get install -y iperf3 libusb-dev
+
+# this is a workaround from uv to enable using vscode launch/python debug
+RUN echo "alias python='uv run python'" >> /home/user/.bashrc
```
NickCao, line 42: A better solution is to configure vscode with a custom interpreter path, as in https://github.com/jumpstarter-dev/jumpstarter/blob/70717aef5824047de67bfa1183bbcd747d178d42/.devcontainer/devcontainer.json#L41

.devfile.yaml
```diff
@@ -0,0 +1,39 @@
+schemaVersion: 2.3.0
+metadata:
+  name: jumpstarter-1hpi
+  namespace: che-kube-admin-devspaces-jhpf17
```
NickCao, line 4: Can we not hardcode these?

.devfile.yaml
```diff
@@ -0,0 +1,39 @@
...
+projects:
+  - name: jumpstarter
+    git:
+      remotes:
+        origin: https://github.com/jumpstarter-dev/jumpstarter.git
+components:
+  - name: runtime
+    container:
+      image: quay.io/jumpstarter-dev/jumpstarter-devspace:latest
```
NickCao, line 13: What would happen if the image is out of sync with the git remote (e.g. working on a different branch other than main)

.devcontainer/Dockerfile
mangelajo: Oh, ok, This is not necessary on this Dockerfile then, I will check if there is an equivalent for devfiles. Thanks

.devfile.yaml
mangelajo: We shouldn't, removing it, thanks.

.devfile.yaml
NickCao: What about using an image component: https://devfile.io/docs/2.3.0/adding-an-image-component

NickCao (approved):

PR #162, opened 2024-10-29
Devspaces client container

mangelajo:


mangelajo:
This container can be consume from client devspaces with the purpose of
creating tests, interacting with jumpstarter, etc.

NickCao:
What's the difference between this container and the client container we are currently shipping? (Apart from the labels)

mangelajo:
> What's the difference between this container and the client container we are currently shipping? (Apart from the labels)

It's based on https://github.com/devfile/developer-images/blob/main/base/ubi9/Dockerfile which is ready-made to run under apache-che / devspaces.

NickCao:
Really hope we have `include` directive in dockerfile :(

NickCao (approved):

mangelajo:
> Really hope we have `include` directive in dockerfile :(

I completely agree!, I've had the same thought a thousand times.

Issue #182, opened 2024-11-25
Exporter runs using 100% CPU of one core


mangelajo:
Sometimes an exporter will enter a cycle where it remains disconnected (probably after a restart of the server that regenerates certs), it pegs 1 CPU core to 100%.

If you strace the process this is found:
```
[pid 630761] <... futex resumed>)       = -1 ETIMEDOUT (Connection timed out)
[pid 630761] futex(0x7f1b92f11b40, FUTEX_WAIT_PRIVATE, 0, {tv_sec=0, tv_nsec=265903915} <unfinished ...>
[pid 630765] <... futex resumed>)       = -1 ETIMEDOUT (Connection timed out)
[pid 630765] futex(0x7f1b95971a90, FUTEX_WAKE_PRIVATE, 1) = 0
[pid 630765] futex(0x7f1b95971a88, FUTEX_WAIT_BITSET_PRIVATE, 0, {tv_sec=2147302, tv_nsec=620982190}, FUTEX_BITSET_MATCH_ANY) = -1 ETIMEDOUT (Connection timed out)
[pid 630765] futex(0x7f1b95971a90, FUTEX_WAKE_PRIVATE, 1) = 0
[pid 630765] futex(0x7f1b95971a88, FUTEX_WAIT_BITSET_PRIVATE, 0, {tv_sec=2147302, tv_nsec=626122564}, FUTEX_BITSET_MATCH_ANY) = -1 ETIMEDOUT (Connection timed out)
[pid 630765] futex(0x7f1b95971a90, FUTEX_WAKE_PRIVATE, 1) = 0
[pid 630765] futex(0x7f1b95971a88, FUTEX_WAIT_BITSET_PRIVATE, 0, {tv_sec=2147302, tv_nsec=631312357}, FUTEX_BITSET_MATCH_ANY) = -1 ETIMEDOUT (Connection timed out)
[pid 630765] futex(0x7f1b95971a90, FUTEX_WAKE_PRIVATE, 1) = 0
[pid 630765] futex(0x7f1b95971a88, FUTEX_WAIT_BITSET_PRIVATE, 0, {tv_sec=2147302, tv_nsec=636480338}, FUTEX_BITSET_MATCH_ANY) = -1 ETIMEDOUT (Connection timed out)
[pid 630765] futex(0x7f1b95971a90, FUTEX_WAKE_PRIVATE, 1) = 0
[...]

PR #183, opened 2024-11-25
Make the exporter exit on error and return error code 1

mangelajo:
Fixes-Issue: #182


mangelajo:
There is a new failure happening on the existing containers builds that built before ':D I'll try to fix that first

kirkbrauer (approved):

PR #185, opened 2024-11-27
Fix Service install docs

kirkbrauer:
- Fix incorrect URL for installing Service CLI
- Fix instructions to use `sudo install` instead of downloading binary into `/usr/local/bin`
- FIx issues with kind and minikube install docs, tested minikube docs to ensure that it works


mangelajo (approved):
Thank you very much Kirk!

PR #187, opened 2024-11-29
FIx typo on "exceptions" when getting connection error

kirkbrauer:


mangelajo:
Thanks, this was also fixed in: https://github.com/jumpstarter-dev/jumpstarter/pull/186/files#diff-577ff2e6ae160e183092cf9c0cffc3b8491c8f521d315729728452989a010829

(xinx!) 🗡️  :-D

PR #196, opened 2024-12-15
introduce tftp driver

bennyz:
Simple TFTP driver to upload/download files to/from a TFTP serve.

Sample config:
```yaml
  tftp:
    type: jumpstarter_driver_tftp.driver.TftpServer
    config:
      root_dir: "/var/lib/tftpboot/"
```


jumpstarter/common/serde.py
```diff
@@ -7,8 +8,20 @@
 
 
 def encode_value(v: Any):
+    if isinstance(v, Generator):
```
bennyz, line 11: Not sure about this

contrib/drivers/tftp/src/jumpstarter_driver_tftp/driver.py
```diff
@@ -0,0 +1,120 @@
...
+class FileNotFound(TftpError):
+    """File not found"""
+    pass
+
+@dataclass(kw_only=True)
+class TftpServer(Driver):
+    """TFTP Server driver for Jumpstarter"""
+
+    root_dir: str
+    host: str = "0.0.0.0"
```
mangelajo, line 35: Some suggestions:

I'd try to reduce configuration surface, (probably setting up a exporter is a cumbersome task, and we can try to make it work with as little tweaks as possible, while allowing for configuration), i.e.:

Make root_dir optional, if not provided, we just pick a TempDir of some sort (or a default /var/lib/jumpstarter/tftp location ? , may be tmp is problematic with ram)

Same for host. We could auto detect the default route IP (which may work on most situations), and optionally accept a specific IP (useful if the DUT is connected over a specific interface which is not the default route interface).

contrib/drivers/tftp/src/jumpstarter_driver_tftp/client.py
```diff
@@ -0,0 +1,49 @@
...
+
+    async def put_file(self, filename: str, data: bytes, chunk_size: int = 1024 * 1024):
+        """Stream file upload using chunked data"""
+        if not data:
+            raise TftpError("Empty data")
+
+        chunks = [data[i:i+chunk_size] for i in range(0, len(data), chunk_size)]
+        total_chunks = len(chunks)
+
+        with tqdm(total=len(data), unit='B', unit_scale=True, desc=f"Uploading {filename}") as pbar:
```
mangelajo, line 28: Nice, progress bars always make waiting nicer :)

contrib/drivers/tftp/src/jumpstarter_driver_tftp/client.py
```diff
@@ -0,0 +1,49 @@
...
+            for i, chunk in enumerate(chunks):
+                await self.call_async(
```
mangelajo, line 30: hehe I was meaning something different for the async here, although it's a starting point, I think we should merge it as it is and we can improve it later.

Where I think async would help in general is the driver side more than the client side

Also using the opendal adapter, I believe (if I am not wrong) it creates a stream (another connection through the router) just for pumping bytes through without waiting (like TCP windows), the advantage of this is that... if even the connection is behind some serious latency the bandwidth will be maximized,

Although, the code is a little bit complex to follow, I think we need to document and clean it up.

 in the current implementation, every chunk needs to receive the exporter side ack before the client can send the new chunk I think.

**But** we can improve that later honestly, it's an optimization.

mangelajo: This is a client example:

https://github.com/jumpstarter-dev/jumpstarter/blob/main/jumpstarter/drivers/storage/client.py#L26

And this is a driver example:

https://github.com/jumpstarter-dev/jumpstarter/blob/main/contrib/drivers/dutlink/jumpstarter_driver_dutlink/driver.py#L131

You'd also avoid the b64 enc/dec of parameters

mangelajo: Link to the resource call on the driver: 
https://github.com/jumpstarter-dev/jumpstarter/blob/main/jumpstarter/driver/base.py#L183

This is what connects it back to the client side stream

mangelajo (commented):
Some comments, very cool to see implementation working so quickly.

Thank you benny!

contrib/drivers/tftp/src/jumpstarter_driver_tftp/client.py
bennyz: now that I use the opendal adapter it doens't look as good[1] (does not consider total size), but I'll look into that separately 

[1] read: 30.9MB [00:06, 4.96MB/s]

contrib/drivers/tftp/src/jumpstarter_driver_tftp/client.py
bennyz: done :)

perhaps it should be made clearer somehow, that the upload/download isn't done with tftp

contrib/drivers/tftp/src/jumpstarter_driver_tftp/driver.py
```diff
@@ -0,0 +1,143 @@
...
+                    async for chunk in src:
+                        await dst.send(chunk)
+
+            return filename
+
+        except Exception as e:
+            raise TftpError(f"Failed to upload file: {str(e)}") from e
+
+    @export
+    async def finalize_file(self, filename: str):
```
mangelajo, line 110: I think this one isn't used anymore.

contrib/drivers/tftp/src/jumpstarter_driver_tftp/driver.py
```diff
@@ -0,0 +1,143 @@
...
+        try:
+            os.remove(os.path.join(self.root_dir, filename))
+        except FileNotFoundError as err:
+            raise FileNotFound(f"File {filename} not found") from err
+        except Exception as e:
+            raise TftpError(f"Failed to delete {filename}") from e
+
+    @export
+    def get_host(self) -> str:
+        return self.host
```
mangelajo, line 138: I'd also add a get_port, so if the exporter configuration is using a different port to 69 it can still be discovered by the test.

I believe that uboot can be tuned to a different destination port with the tftpdstp env var.

contrib/drivers/tftp/src/jumpstarter_driver_tftp/driver.py
```diff
@@ -0,0 +1,143 @@
...
+        logger.warning("Could not determine default IP address, falling back to 0.0.0.0")
+        return "0.0.0.0"
+
+class TftpError(Exception):
+    """Base exception for TFTP server errors"""
+    pass
+
+class ServerAlreadyRunning(TftpError):
+    """Server is already running"""
+    pass
```
mangelajo, line 34: nice exception structuring, which I think we have been lacking in some areas

I would remove this one, we don't use it, and two calls to start may be idempotent. (i.e. one test starts, the next just finds it started but could potentially call to start again, nothing should fail)

contrib/drivers/tftp/src/jumpstarter_driver_tftp/client.py
mangelajo: Ah, yes, that may need improvement, the progress bar is a bit meeh

contrib/drivers/tftp/src/jumpstarter_driver_tftp/client.py
mangelajo: right, it may look like we are using put via tftp.

I'd start documenting with docstrings , then it's very easy to expose it in docs.jumpstarter.dev

i.e. https://docs.jumpstarter.dev/api-reference/contrib/ustreamer.html

(From https://github.com/jumpstarter-dev/jumpstarter/tree/main/docs/source/api-reference and https://raw.githubusercontent.com/jumpstarter-dev/jumpstarter/refs/heads/main/docs/source/api-reference/contrib/ustreamer.md )

mangelajo (changes requested):
It looks great, just a couple of comments, this already doesn't feel like draft ;)

contrib/drivers/tftp/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM fedora:40 AS builder
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+RUN --mount=target=/src uv build --project /src --out-dir /dist
+
+FROM quay.io/jumpstarter-dev/jumpstarter:latest
+RUN --mount=from=builder,source=/dist,target=/dist \
+  VIRTUAL_ENV=/jumpstarter uv pip install /dist/*.whl
```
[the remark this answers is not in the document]
bennyz: yeah, I took everything from the template, I'll remove the stuff that isn't in other drivers in `contrib`

contrib/drivers/tftp/src/jumpstarter_driver_tftp/driver.py
mangelajo: you forgot about get_port

Everything else looks great.

mangelajo (changes requested):
one tiny nit, everything else looks great, I guess from our chat still we need to figure out the timeouts.

contrib/drivers/tftp/src/jumpstarter_driver_tftp/driver.py
bennyz: it's on L#119, or did you mean something else?

mangelajo: The one on L174 actually :) I must have been running the last time.

contrib/drivers/tftp/src/jumpstarter_driver_tftp/server_test.py
```diff
@@ -0,0 +1,173 @@
...
+
+    server_task = asyncio.create_task(server.start())
+    await asyncio.sleep(0.1)
+
+    try:
+        transport, _ = await create_test_client(server.port)
+
+        rrq_packet = (
+            Opcode.RRQ.to_bytes(2, 'big') +
+            b'../../../etc/passwd\x00' +
```
mangelajo, line 161: 👍

mangelajo (approved):

mangelajo:
@bennyz there is one ruff error, and if possible squash the commits, as the current merge queue doesn't seem to accept squash merging.

mangelajo (approved):

PR #206, opened 2024-12-26
Handle all instances of tls credentials

mangelajo:
The patch handling tls configuration on exporter/client didn't pass down the tls configuration to the router connections, so when set to insecure, control connection would work, but router connections failed.

Related-Issue: #192


mangelajo:
oops I created a circular import :-]

bennyz (approved):

PR #210, opened 2025-01-05
http: simplify test and fix minor issues

bennyz:
- simplify tests, and add a test for file retrieval
- remove unused loop param
- call shutdown before cleanup, as docs require

### DIFF
--- a/contrib/drivers/http/src/jumpstarter_driver_http/driver_test.py
+++ b/contrib/drivers/http/src/jumpstarter_driver_http/driver_test.py
@@ -1,34 +1,78 @@
-from pathlib import Path
+import os
+import uuid
 from tempfile import TemporaryDirectory
 
+import aiohttp
+import anyio
 import pytest
+from anyio import create_memory_object_stream
 
-from jumpstarter.common.utils import serve
+from jumpstarter.common.resources import ClientStreamResource
 
 from .driver import HttpServer
 
 
-@pytest.mark.asyncio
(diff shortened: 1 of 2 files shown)


contrib/drivers/http/src/jumpstarter_driver_http/driver_test.py
```diff
@@ -1,34 +1,78 @@
...
+    test_content = b"test content"
 
-            deleted_filename = client.delete_file("test.txt")
-            assert deleted_filename == "test.txt"
+    send_stream, receive_stream = create_memory_object_stream(max_buffer_size=1024)
 
-            files_after_deletion = client.list_files()
-            assert "test.txt" not in files_after_deletion
+    resource_uuid = uuid.uuid4()
+    server.resources[resource_uuid] = receive_stream
```
mangelajo, line 41: Hi, nice work,

Although I'd want us to consider if we still want to keep some test through the serve() / write_local_file method too.

The reason to propose it is that we otherwise stop testing the client slide at all, and all the inner interactions of client->exporter

bennyz: yeah, I figured I'd make it more similar to the tftp driver test and we can perhaps have a seperate test for the client, because in my initial attempt to do the HTTP GET in the context of `serve` didn't go well because of the mixed async/sync. 
However, I will try again, perhaps it's simpler than it looks

PR #212, opened 2025-01-08
Fix JMP_LEASE handling in pytest

mangelajo:
1) Import the right module for the env_var name
2) Only release runtime created leases, leases received via JMP_LEASE
   are kept, because those must be manually released.

### DIFF
--- a/jumpstarter/testing/pytest.py
+++ b/jumpstarter/testing/pytest.py
@@ -1,5 +1,4 @@
 import logging
-import os
 import time
 from typing import ClassVar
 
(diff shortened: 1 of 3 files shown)


jumpstarter/testing/pytest.py
```diff
@@ -59,10 +60,14 @@ def client(self):
         except RuntimeError:
             labels = getattr(self, "filter_labels", {})
             config = ClientConfigV1Alpha1.load("default")
-            lease_name = os.getenv(env.JMP_LEASE, None)
+            lease_name = os.getenv(env_vars.JMP_LEASE, None)
+            release = True
```
bennyz, line 64: not important at all but this is similar to:
https://github.com/jumpstarter-dev/jumpstarter/blob/main/jumpstarter/cli/client.py#L227

jumpstarter/testing/pytest.py
```diff
@@ -7,6 +7,7 @@
 
 from jumpstarter.common import MetadataFilter
 from jumpstarter.common.utils import env
+from jumpstarter.config import env as env_vars
```
bennyz, line 10: python 😅

mangelajo: yup ':D I am surprised that the linter didn't catch it.

jumpstarter/testing/pytest.py
mangelajo: true, and the other code is better, 

We should probably make this part of the config.lease logic instead.

If lease_name=None , release is assumed to be True, if lease_name is provided ... then assume it to be False.

bennyz (approved):

PR #223, opened 2025-01-21
Init DutlinkSerial driver

NickCao:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new class for enhanced serial configuration handling.
- **Refactor**
	- Improved code structure for serial connection management.
	- Minor formatting improvements in logging and error handling.

### DIFF
--- a/contrib/drivers/dutlink/jumpstarter_driver_dutlink/driver.py
+++ b/contrib/drivers/dutlink/jumpstarter_driver_dutlink/driver.py
@@ -83,6 +83,22 @@ class DutlinkConfig:
 
 
 @dataclass(kw_only=True)
+class DutlinkSerial(DutlinkConfig, PySerial):
+    url: str | None = field(init=False, default=None)
+    alternate_console: str | None = field(default=None)
+
+    def __post_init__(self):
+        super().__post_init__()
+
+        if self.alternate_console is not None:
+            self.url = self.alternate_console
+        else:
+            self.url = self.tty
+
+        super(PySerial, self).__post_init__()
+
+
+@dataclass(kw_only=True)
 class DutlinkPower(DutlinkConfig, PowerInterface, Driver):
     last_action: str | None = field(default=None)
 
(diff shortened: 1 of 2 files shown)


contrib/drivers/dutlink/jumpstarter_driver_dutlink/driver.py
```diff
@@ -82,6 +82,22 @@ def control(self, direction, ty, actions, action, value):
             return str_value
 
 
+@dataclass(kw_only=True)
+class DutlinkSerial(DutlinkConfig, PySerial):
+    url: str | None = field(init=False, default=None)
```
mangelajo, line 87: Is it necessary to re-declare the `url`   field from PySerial?

contrib/drivers/dutlink/jumpstarter_driver_dutlink/driver.py
```diff
@@ -82,6 +82,22 @@ def control(self, direction, ty, actions, action, value):
...
+    alternate_console: str | None = field(default=None)
```
mangelajo, line 88:
```suggestion
    alternate_console: str | None = field(default=None)
```

I think in this case the alternate console is not necessary anymore. If you are going to use an alternate console standalone, you should just use pyserial.

The alternate console is for some devices that will expose the DUT console as a USB port that comes from the DUT itself. I would just open the console found for the specific serial number, and that's it.

TL;DR

i.e. you can see special handling here https://github.com/jumpstarter-dev/jumpstarter/pull/223/files#diff-a1e61e9045a78ae1ee5d8ac8863402c4dfb5237c04fa246119da0c3fe3df8e64R269 because in some cases the DUT has to be powered on for the console to show up, if anybody needs that type of handling should go for the composite DUTLink and not care about this.

contrib/drivers/dutlink/jumpstarter_driver_dutlink/driver.py
NickCao: Yes, otherwise init would fail with `TypeError: DutlinkSerial.__init__() missing 1 required keyword-only argument: 'url'`

mangelajo (approved):
Thank you, looks great!

PR #227, opened 2025-01-22
Rework gpio driver with adafruit-blinka

NickCao:


NickCao:
Since https://github.com/adafruit/Adafruit_Blinka/commit/b277397feff980a3f879538a1243e6f7dba7b2d2, sysv_ipc is installed unconditionally, which is very annoying since it has c extensions to be built from source.

mangelajo:
Looks great!

We need to add gcc to the container images building the wheels,  (see the container build error)

also, awesome if we can include documentation https://github.com/jumpstarter-dev/jumpstarter/tree/main/docs/source/api-reference on the PR :)

Thank you!

mangelajo:
> Since [adafruit/Adafruit_Blinka@b277397](https://github.com/adafruit/Adafruit_Blinka/commit/b277397feff980a3f879538a1243e6f7dba7b2d2), sysv_ipc is installed unconditionally, which is very annoying since it has c extensions to be built from source.

let's just add gcc to the builds and let's live with it.

NickCao:
> let's just add gcc to the builds and let's live with it.

Problem is sysv_ipc is built when installing the wheels, not when building the wheels, so it would be part of the final product.

mangelajo:
> > let's just add gcc to the builds and let's live with it.
> 
> Problem is sysv_ipc is built when installing the wheels, not when building the wheels, so it would be part of the final product.

Ohhh, right!, thanks for explaining

It is the same problem as with gRPC, i.e. for RV64 it did not have precompiled wheels and it needed to compile on install, but they have pre-compiled wheels for most architectures though...

NickCao:
Let's leave gcc there and call it a day, will surely find a solution in the long term.

NickCao:
> also, awesome if we can include documentation https://github.com/jumpstarter-dev/jumpstarter/tree/main/docs/source/api-reference on the PR :)

Documented DigitalIOClient

docs/source/api-reference/drivers/raspberrypi.md
```diff
@@ -0,0 +1,8 @@
+# Raspberry Pi drivers
+
```
mangelajo, line 2: can we extend it in a similar way to what we have in pyserial, so that we provide examples of exporter configuration, as well as examples of client (python and shell - for the power version -)

NickCao:
Rebased on current main.

mangelajo:
Thanks, testing it, will merge once validated on a rpi4

mangelajo:
Closing in favor of https://github.com/jumpstarter-dev/jumpstarter/pull/544

PR #230, opened 2025-01-23
fix and expand dutlink tests

NickCao:


mangelajo (dismissed):

mangelajo (commented):
Should we keep the specific exception capture?


Thanks for doing this, I have a dutlink prepared on the lab to hook up to these tests

mangelajo (approved):

PR #231, opened 2025-01-24
Introducing rust to jumpstarter

NickCao:


contrib/libs/jumpstarter-rust/proto/jumpstarter/v1/jumpstarter.proto
```diff
@@ -0,0 +1,219 @@
+// Copyright 2024 The Jumpstarter Authors
```
mangelajo, line 1: may be we can use a git submodule here ;)

NickCao: Or just use buf.build as with python.

mangelajo:
The idea of rust exporters (without python I think is interesting), eventually useful to run on the DUT, or for some memory constrained exporters, or exporters without python..


But IMHO adding the dependency to rust for the python side ... makes it more complicated to maintain/package/etc, we should focus on priorities that will help us bring this project UP to stability and limit complexity.

NickCao:
This is not happening now.

PR #234, opened 2025-01-26
Fix Ruff version mismatch and config issues

kirkbrauer:
## Summary by CodeRabbit

- **Chores**
	- Updated GitHub Actions workflow configuration with minor capitalization changes.
	- Updated pre-commit configuration to use newer Ruff version (v0.9.2).
	- Updated Ruff dependency to exact version 0.9.2 in project configuration.
	- Modified Dockerfile to streamline the build process by using pre-built binaries.
- **New Features**
	- Introduced a new command `request` to the `lease` command group for requesting an exporter lease.
- **Bug Fixes**
	- Removed unnecessary Ruff dependency from development groups in multiple project configurations.

### DIFF
--- a/packages/jumpstarter-cli-client/jumpstarter_cli_client/lease.py
+++ b/packages/jumpstarter-cli-client/jumpstarter_cli_client/lease.py
@@ -1,11 +1,10 @@
 import asyncclick as click
-from jumpstarter_cli_common import AliasedGroup
-
 from jumpstarter.common import MetadataFilter
 from jumpstarter.config import (
     ClientConfigV1Alpha1,
     UserConfigV1Alpha1,
 )
+from jumpstarter_cli_common import AliasedGroup
 
 
 @click.group(cls=AliasedGroup, short_help="")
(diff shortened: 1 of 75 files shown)


packages/jumpstarter-cli-client/jumpstarter_cli_client/lease.py
```diff
@@ -47,31 +47,32 @@ def lease_release(name, lease, all_leases):
...
 
-.. code-block:: bash
+    .. code-block:: bash
 
-    $ JMP_LEASE=$(jmp lease request -l label match)
-    $ jmp shell
-    $$ j --help
-    $$ exit
-    $ jmp lease release -l "${JMP_LEASE}"
+        $ JMP_LEASE=$(jmp lease request -l label match)
```
mangelajo, line 69: We need to update to jmp client lease request  

But the important thing is fixing ruff now... merging :)

mangelajo (approved):

packages/jumpstarter-cli-client/jumpstarter_cli_client/lease.py
kirkbrauer: I am a bit worried about the non deterministic behavior of ruff. Maybe we need to make sure everyone upgrades their uv and fully syncs the workspace. 🤷‍♂️

PR #241, opened 2025-01-29
tftp: add checksum validation

bennyz:
## Summary by CodeRabbit

- **New Features**
	- Added checksum verification for TFTP file uploads.
	- Implemented file integrity checks to prevent redundant uploads.
	- Introduced a new constant for configurable chunk size.

- **Bug Fixes**
	- Improved file management with checksum tracking.
	- Enhanced error handling for TFTP operations.

- **Tests**
	- Updated test suite to validate new checksum functionality.
	- Added comprehensive tests for file upload and integrity checks.
	- Introduced tests for detecting corrupted files and handling different checksums.
	- Enhanced modularity in test functions for better maintainability.
	- Streamlined file upload process in tests with a new helper function.

- **Documentation**
	- Updated configuration options for testing in `pyproject.toml`.


packages/jumpstarter-driver-tftp/jumpstarter_driver_tftp/client.py
```diff
@@ -46,37 +49,30 @@ def list_files(self) -> list[str]:
...
+            logger.info(f"Skipping upload of identical file: {filename}")
+            return filename
 
-        Returns:
-            str: Name of the uploaded file
-        """
-        filename = Path(path).name
         with OpendalAdapter(client=self, operator=operator, path=path, mode="rb") as handle:
-            return self.call("put_file", filename, handle)
+            return self.call("put_file", filename, handle, client_checksum)
```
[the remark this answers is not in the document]
bennyz: @coderabbit, is this actually necessary? look at driver.py and its tests

packages/jumpstarter-driver-tftp/jumpstarter_driver_tftp/client.py
```diff
@@ -108,3 +104,10 @@ def get_port(self) -> int:
             int: The port number (default is 69)
         """
         return self.call("get_port")
+
+    def _compute_checksum(self, operator: Operator, path: str) -> str:
+        hasher = hashlib.sha256()
+        with operator.open(path, "rb") as f:
+            while chunk := f.read(8192):
```
mangelajo, line 111:
```suggestion
            while chunk := f.read(8192):
```

Probably even go for bigger chunks, it will work much quicker. 1MB, 4MB ... etc.

Can we add some debug logging before, after, printing "we are starting the checksum of  blahblah", "the checksum of blahblah is xxxxxxxxxxxxxxxx"

In such way we can identify how long does it take to compute in debug mode.

packages/jumpstarter-driver-tftp/jumpstarter_driver_tftp/driver.py
```diff
@@ -184,3 +215,45 @@ def close(self):
...
+            checksum_path = self._get_checksum_path(filename)
+            with open(checksum_path, 'w') as f:
+                f.write(f"{checksum}\n")
+        except Exception as e:
+            logger.error(f"Failed to write checksum file for {filename}: {e}")
+
+    def _compute_checksum(self, path: str) -> str:
+        hasher = hashlib.sha256()
+        with open(path, "rb") as f:
+            while chunk := f.read(8192):
```
mangelajo, line 244: same about the block size.

packages/jumpstarter-driver-tftp/jumpstarter_driver_tftp/driver.py
```diff
@@ -184,3 +215,45 @@ def close(self):
...
+        for filename in os.listdir(self.root_dir):
+            if filename.endswith(self.checksum_suffix):
+                continue
+            file_path = os.path.join(self.root_dir, filename)
+            if os.path.isfile(file_path):
+                stored_checksum = self._read_checksum_file(filename)
+                current_checksum = self._compute_checksum(file_path)
+                if stored_checksum != current_checksum:
+                    self._write_checksum_file(filename, current_checksum)
+                self._checksums[filename] = current_checksum
```
mangelajo, line 259: I would not initialize it for existing files. I.e. if we connect this path to an existing NFS, it's going to be slow. I would compute checksum:
1) At write time...
2) At checksum check time, if the checksum did not exist, but then we cache it.

packages/jumpstarter-driver-tftp/jumpstarter_driver_tftp/driver.py
```diff
@@ -184,3 +215,45 @@ def close(self):
         if self.server_thread is not None:
             self.stop()
         super().close()
+
+    def _get_checksum_path(self, filename: str) -> str:
+        return os.path.join(self.root_dir, f"{filename}{self.checksum_suffix}")
+
+    def _read_checksum_file(self, filename: str) -> Optional[str]:
+        try:
+            checksum_path = self._get_checksum_path(filename)
```
mangelajo, line 224: We should verify that the checksum cache file is older than the file.

Otherwise it would mean that the file has been updated (out of our control) and that we need to regenerate the checksum at the right time.

mangelajo (changes requested):
Also keep in mind that we may want to generalize this checksum thing to other methods: http, tftp, iscsi, etc...

packages/jumpstarter-driver-tftp/jumpstarter_driver_tftp/driver.py
bennyz: still need to address this one

bennyz:
decided to drop caching for now, we need something more sophisticated to handle invalidation and the performance gain isn't expected to be very significant given we aren't expected to recompute the checksum on the same file very often

mangelajo:
> decided to drop caching for now, we need something more sophisticated to handle invalidation and the performance gain isn't expected to be very significant given we aren't expected to recompute the checksum on the same file very often

Ack, makes sense, we can optimize anytime later if we really need to.

mangelajo:
Looks good, can you also open a PR documenting TFTP like this: 

https://docs.jumpstarter.dev/api-reference/drivers/pyserial.html

?

mangelajo (approved):

PR #251, opened 2025-02-05
snmp: initial SNMP driver for power control

bennyz:
Simple SNMP driver for power management

```yaml
apiVersion: jumpstarter.dev/v1alpha1
kind: ExporterConfig
endpoint: grpc.jumpstarter.apps.my.cluster.com:443
tls:
  ca: ''
  insecure: true
token: <token>
export:
  power:
    type: "jumpstarter_driver_snmp.driver.SNMPServer"
    config:
      host: "pdu.mgmt.com"
      user: "labuser"
      plug: 32
      quiescent_period: 3
      oid: "1.3.6.1.4.1.13742.6.4.1.2.1.2.1"
```

## Summary by CodeRabbit

- **New Features**
  - Introduced a new client interface for SNMP power control, allowing users to manage power states and access a command-line interface.
  - Added support for power cycling functionality through a new method in the SNMP driver.
  - Added a new configuration resource for the SNMP driver with necessary parameters for SNMP communication.

- **Documentation**
  - Enhanced documentation with configuration examples and API usage for the SNMP driver, including detailed descriptions of parameters and methods.
  - Included the SNMP driver in the documentation table of contents.

- **Chores**
  - Updated project configuration files to include new dependencies and project metadata for the SNMP driver.


bennyz:
@coderabbitai review

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver.py
```diff
@@ -0,0 +1,166 @@
...
+            snmp_engine,
+            udp.DOMAIN_NAME,
+            udp.UdpAsyncioTransport().open_client_mode()
+        )
+
+        config.add_target_address(
+            snmp_engine,
+            "my-target",
+            udp.DOMAIN_NAME,
+            (self.ip_address, self.port),
```
NickCao, line 80: Can't `pysnmp` accept a domain name as the target address? The config option itself is named `DOMAIN_NAME`.

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/client.py
```diff
@@ -0,0 +1,48 @@
+from dataclasses import dataclass
+
+import asyncclick as click
+
+from jumpstarter.client import DriverClient
+
+
+@dataclass(kw_only=True)
+class SNMPServerClient(DriverClient):
```
NickCao, line 9: We can inherit from `PowerClient`

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver.py
bennyz: I got `OSError: illegal IP address string passed to inet_pton`, but didn't look too deep to see if there's a way without manual resolve

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/client.py
bennyz: ack

bennyz:
todo:
- [ ] inherit PowerClient
- [ ] add tests
- [ ] add examples
- [ ] add docs

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/client.py
```diff
@@ -0,0 +1,48 @@
...
+    """Client interface for SNMP Power Control"""
+
+    def power_on(self):
```
mangelajo, line 12: yes, let's use standard on(), off() , ... and a fake power read.

power_cycle should be implemented on the client side, we can add it to the base `PowerClient` class (may be with some parameters like the power off time, and a default of 1 or 2 s?)

I would just call it cycle() (so it's aligned with on/off)

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/client.py
```diff
@@ -0,0 +1,48 @@
...
+        def off():
+            """Turn power off"""
+            result = self.power_off()
+            click.echo(result)
+
+        @snmp.command()
+        def cycle():
+            """Power cycle the device"""
+            result = self.power_cycle()
+            click.echo(result)
```
mangelajo, line 46: And then we can also add this bit to the base Power Interface.


In fact, if you do this, there is no need to create a Client, we can just reference the base power client from the driver itself.

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver.py
```diff
@@ -0,0 +1,166 @@
...
+    plug: int = field()
+    oid: str = field(default="1.3.6.1.4.1.13742.6.4.1.2.1.2.1")
+
+    def __post_init__(self):
+        if hasattr(super(), "__post_init__"):
+            super().__post_init__()
+
+        try:
+            self.ip_address = socket.gethostbyname(self.host)
+            self.logger.info(f"Resolved {self.host} to {self.ip_address}")
```
mangelajo, line 40:
```suggestion
            self.logger.debug(f"Resolved {self.host} to {self.ip_address}")
```

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver.py
```diff
@@ -0,0 +1,166 @@
...
+            asyncio.set_event_loop(loop)
+
+
+        snmp_engine = engine.SnmpEngine()
+
+        config.add_v3_user(
+            snmp_engine,
+            self.user,
+            config.USM_AUTH_NONE,
+            None
```
mangelajo, line 61: I would add some ability to authenticate too.

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver.py
```diff
@@ -0,0 +1,166 @@
...
+    @classmethod
+    def client(cls) -> str:
+        return "jumpstarter_driver_snmp.client.SNMPServerClient"
+
+    def _snmp_set(self, state: PowerState):
+        result = {"success": False, "error": None}
+
+        def callback(snmpEngine, sendRequestHandle, errorIndication,
+                    errorStatus, errorIndex, varBinds, cbCtx):
+            self.logger.info(f"Callback {errorIndication} {errorStatus} {errorIndex} {varBinds}")
```
mangelajo, line 95:
```suggestion
            self.logger.debug(f"Callback {errorIndication} {errorStatus} {errorIndex} {varBinds}")
```

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver.py
```diff
@@ -0,0 +1,166 @@
...
+            if errorIndication:
+                self.logger.info(f"SNMP error: {errorIndication}")
```
mangelajo, line 97:
```suggestion
                self.logger.error(f"SNMP error: {errorIndication}")
```

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver.py
```diff
@@ -0,0 +1,166 @@
...
+                result["error"] = f"SNMP error: {errorIndication}"
+            elif errorStatus:
+                self.logger.info(f"SNMP status: {errorStatus}")
```
mangelajo, line 100:
```suggestion
                self.logger.error(f"SNMP status: {errorStatus}")
```

? what is the difference between errorStatus and errorIndication?

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver.py
```diff
@@ -0,0 +1,166 @@
...
+                result["error"] = (
+                    f"SNMP error: {errorStatus.prettyPrint()} at "
+                    f"{varBinds[int(errorIndex) - 1][0] if errorIndex else '?'}"
+                )
+            else:
+                result["success"] = True
+                for oid, val in varBinds:
+                    self.logger.debug(f"{oid.prettyPrint()} = {val.prettyPrint()}")
+            self.logger.info(f"SNMP set result: {result}")
```
mangelajo, line 109:
```suggestion
            self.logger.debug(f"SNMP set result: {result}")
```

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver.py
```diff
@@ -0,0 +1,166 @@
...
+
+            return f"Power {state.name} command sent successfully"
+
+        except Exception as e:
+            error_msg = f"SNMP set failed: {str(e)}"
+            self.logger.error(error_msg)
+            raise SNMPError(error_msg) from e
+
+    @export
+    def power_on(self):
```
mangelajo, line 139:
```suggestion
    def on(self):
```

So you can use the default power interface. https://github.com/jumpstarter-dev/jumpstarter/pull/249/files#diff-a6486ddc66753eb801b3d8b638ebe912ae37c36fb9e6d773e5850316a7b77058R42

I need to update my base class in there anyway

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver.py
```diff
@@ -0,0 +1,166 @@
...
+        """Turn power on"""
+        return self._snmp_set(PowerState.ON)
+
+    @export
+    def power_off(self):
```
mangelajo, line 144:
```suggestion
    def off(self):
```

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver.py
```diff
@@ -0,0 +1,166 @@
...
+            self.logger.info("Starting power cycle sequence")
+            self.power_off()
+            self.logger.info(f"Waiting {self.quiescent_period} seconds...")
+            time.sleep(self.quiescent_period)
+            self.power_on()
+            return "Power cycle completed successfully"
+        except Exception as e:
+            error_msg = f"Power cycle failed: {str(e)}"
+            self.logger.error(error_msg)
+            raise SNMPError(error_msg) from e
```
mangelajo, line 161:
```suggestion
    def power_cycle(self):
        """Power cycle the device"""
        try:
            self.logger.info("Starting power cycle sequence")
            self.power_off()
            self.logger.info(f"Waiting {self.quiescent_period} seconds...")
            time.sleep(self.quiescent_period)
            self.power_on()
            return "Power cycle completed successfully"
        except Exception as e:
            error_msg = f"Power cycle failed: {str(e)}"
            self.logger.error(error_msg)
            raise SNMPError(error_msg) from e
```

Let's move this as a client side implementation to the jumpstarter_driver_power.client.PowerClient.

mangelajo (changes requested):

mangelajo:
Great work, don't get discouraged by the amount of comments, it's only about standardizing our interfaces as much as possible. Then when you have some implementation it really doesn't matter what is the underlaying power driver or storage system/write method, etc...

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver.py
bennyz: not entirely sure, but looking at https://github.com/lextudio/pysnmp/blob/ba05fad934b3b904e02d3ceb7f1ca9c1b332560f/pysnmp/proto/api/v1.py#L191

it seems error status is more about something that comes from the PDU, while indication is more about var binds processing

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver.py
```diff
@@ -0,0 +1,179 @@
...
+class SNMPServer(Driver):
+    """SNMP Power Control Driver"""
+    host: str = field()
+    user: str = field()
+    port: int = field(default=161)
+    quiescent_period: int = field(default=5)
+    timeout: int = 3
+    plug: int = field()
+    oid: str = field(default="1.3.6.1.4.1.13742.6.4.1.2.1.2.1")
+    auth_protocol: str = field(default=None)  # 'MD5' or 'SHA'
```
bennyz, line 32: need to change these to an enum

packages/jumpstarter-driver-power/jumpstarter_driver_power/client.py
```diff
@@ -13,6 +14,16 @@ def on(self) -> str:
     def off(self) -> str:
         return self.call("off")
 
+    def cycle(self, quiescent_period: int = 2) -> str:
```
mangelajo, line 17:
```suggestion
    def cycle(self, quiescent_period: int = 2):
```

I would simply return nothing (i.e. avoid issues later if somebody is using a linter, and the linter complains because you are not capturing the output of the str... although I see we have the same on on/off hhmmmm

WDYT @NickCao ?

docs/source/api-reference/drivers/snmp.md
```diff
@@ -0,0 +1,82 @@
...
+Returns:
+- str: Confirmation message
+
+#### off()
+Turn power off for the configured outlet.
+
+Returns:
+- str: Confirmation message
+
+#### cycle(quiescent_period: int = 2)
```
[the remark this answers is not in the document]
bennyz: I actually should change it wait, or hold, I only used quiescent_period because it's the term used in beaker

packages/jumpstarter-driver-power/jumpstarter_driver_power/client.py
NickCao: > I would simply return nothing

Sure, the return value was initially added for no particular reasons. We should also make `on` and `off` return nothing.

bennyz: ack, i'll drop it

bennyz: had to change a lot of tests a result 😅

mangelajo: oops, sorry for the extra noise :D we could have done it later, thanks benny!

packages/jumpstarter-driver-power/jumpstarter_driver_power/client_test.py
```diff
@@ -5,9 +5,6 @@
 
 def test_client_mock_power():
     with serve(MockPower()) as client:
-        assert client.on() == "ok"
-        assert client.off() == "ok"
-
```
mangelajo, line 10: I wouldn't fully kill the calls 

at least call client.on(I)
client.off() as a smoke test :) (no exceptions, etc..)

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver.py
```diff
@@ -0,0 +1,203 @@
...
+            self.logger.debug(f"SNMP set result: {result}")
+
+        try:
+            self.logger.info(f"Sending power {state.name} command to {self.host}")
+            try:
+                asyncio.get_running_loop()
+            except RuntimeError:
+                loop = asyncio.new_event_loop()
+                asyncio.set_event_loop(loop)
+
```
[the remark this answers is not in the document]
mangelajo: What kind of problems can you run on with our implementation?

mangelajo (approved):
just a small thing, ok to follow up

packages/jumpstarter-testing/jumpstarter_testing/pytest_test.py
```diff
@@ -12,7 +12,7 @@ def test_env(pytester: Pytester, monkeypatch):
 
         class TestSample(JumpstarterTest):
             def test_simple(self, client):
-                assert client.on() == "ok"
+                client.on()
```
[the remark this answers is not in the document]
bennyz: good idea actually

packages/jumpstarter-driver-composite/jumpstarter_driver_composite/driver_test.py
```diff
@@ -1,21 +1,23 @@
...
+    child2 = SimpleDriver()
+
+    composite = Composite(children={
+        "child1": child1,
+        "child2": child2
+    })
+
+    assert len(composite.children) == 2
+    assert composite.children["child1"] == child1
+    assert composite.children["child2"] == child2
```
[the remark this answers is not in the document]
bennyz: yes

packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver.py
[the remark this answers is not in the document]
bennyz:
```
asyncio.get_running_loop()
Return the running event loop in the current OS thread.

Raise a [RuntimeError](https://docs.python.org/3/library/exceptions.html#RuntimeError) if there is no running event loop.

This function can only be called from a coroutine or a callback.
```

your approach would cause an exception no?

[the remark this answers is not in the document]
bennyz:
```
asyncio.get_running_loop()
Return the running event loop in the current OS thread.

Raise a [RuntimeError](https://docs.python.org/3/library/exceptions.html#RuntimeError) if there is no running event loop.

This function can only be called from a coroutine or a callback.
```

@coderabbitai  your approach would cause an exception no?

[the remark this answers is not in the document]
bennyz: @coderabbitai  perhaps i do need to close the loop though?

PR #282, opened 2025-02-14
Implement OIDC login

NickCao:
## Summary by CodeRabbit

- **New Features**
  - Introduced interactive login commands to the Jumpstarter CLI tools, enabling a seamless and secure authentication experience.
  - Enhanced OpenID Connect support now allows users to choose between multiple authentication flows for a smoother login process.
  - Added `client_login` and `exporter_login` commands for logging into Jumpstarter instances.
  
- **Dependencies**
  - Added new dependencies for improved authentication and token handling capabilities.

### DIFF
--- a/packages/jumpstarter-cli-client/jumpstarter_cli_client/client_login.py
+++ b/packages/jumpstarter-cli-client/jumpstarter_cli_client/client_login.py
@@ -1,0 +1,103 @@
+import asyncclick as click
+from jumpstarter_cli_common.oidc import Config, decode_jwt_issuer, opt_client_id
+
+from jumpstarter.config import ClientConfigV1Alpha1, ClientConfigV1Alpha1Drivers, ObjectMeta, UserConfigV1Alpha1
+
+
+@click.command("login", short_help="Login")
+@click.argument("alias", default=None)
+@click.option(
+    "-e",
+    "--endpoint",
+    type=str,
+    help="Enter the Jumpstarter service endpoint.",
+    default=None,
+)
+@click.option(
+    "--namespace",
+    type=str,
+    help="Enter the Jumpstarter client namespace.",
+    default=None,
+)
+@click.option(
+    "--name",
+    type=str,
+    help="Enter the Jumpstarter client name.",
+    default=None,
+)
+@click.option("--username", type=str, help="Enter the OIDC username.", default=None)
+@click.option("--password", type=str, help="Enter the OIDC password.", default=None)
+@click.option(
+    "--issuer",
+    type=str,
+    help="Enter the OIDC issuer.",
+    default=None,
+)
+@click.option(
+    "--allow",
+    type=str,
+    help="A comma-separated list of driver client packages to load.",
+    default="",
+)
+@click.option(
+    "--unsafe", is_flag=True, help="Should all driver client packages be allowed to load (UNSAFE!).", default=None
+)
+@opt_client_id
+async def client_login(  # noqa: C901
+    alias: str,
+    endpoint: str,
+    namespace: str,
+    name: str,
+    username: str | None,
+    password: str | None,
+    issuer: str,
+    client_id: str,
+    allow: str,
+    unsafe: str,
+):
+    """Login into a jumpstarter instance"""
+
+    if alias is not None:
+        try:
+            config = ClientConfigV1Alpha1.load(alias)
+            issuer = decode_jwt_issuer(config.token)
+        except FileNotFoundError:
+            if namespace is None:
+                namespace = click.prompt("Enter the Jumpstarter client namespace")
+            if name is None:
+                name = click.prompt("Enter the Jumpstarter client name")
+            if endpoint is None:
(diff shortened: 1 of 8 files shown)


packages/jumpstarter-cli-client/jumpstarter_cli_client/client_login.py
```diff
@@ -0,0 +1,70 @@
...
+    endpoint: str,
+    namespace: str,
+    name: str,
+    username: str | None,
+    password: str | None,
+    issuer: str,
+    allow: str,
+    unsafe: str,
+):
+    """Login into a jumpstarter instance"""
```
mangelajo, line 54: Thoughts:

If the config already exists, we should probably get the details (name, namespace, endpoint, unsafe, allow) from such config instead.

Should we store the issuer information in the config file?, i.e. to allow updating the token when it expires.

packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -0,0 +1,70 @@
+import ssl
+from dataclasses import dataclass
+from typing import ClassVar
+
+import aiohttp
+import truststore
+from aiohttp import web
+from anyio import create_memory_object_stream
+from authlib.integrations.httpx_client import AsyncOAuth2Client
```
mangelajo, line 9: Is it possible to do the cli sync and use the requests library to avoid the extra dependency?

packages/jumpstarter-cli-client/jumpstarter_cli_client/client_login.py
NickCao: I feel like making token a json object including `issuer`, `access_token` and `refresh_token`, so we can handle everything automatically in the background.

packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
NickCao: Done

NickCao:
TODO:
- [ ] make client id configurable
- [ ] make redirect uri random/configurable
- [ ] implement updating/refreshing existing token

packages/jumpstarter-cli-client/jumpstarter_cli_client/client_login.py
NickCao: On wait, two things:
1. since the cli would be a public oidc client, it cannot keep the secrecy of the "client secret", it would not be issued a refresh token anyway.
2. the issuer information is already part of the access_token (a jwt)

NickCao:
Please review but do not merge, the last commit should be dropped before merging.

packages/jumpstarter-cli-client/jumpstarter_cli_client/client_login.py
```diff
@@ -0,0 +1,103 @@
...
+    client_id: str,
+    allow: str,
+    unsafe: str,
+):
+    """Login into a jumpstarter instance"""
+
+    if alias is not None:
+        try:
+            config = ClientConfigV1Alpha1.load(alias)
+            issuer = decode_jwt_issuer(config.token)
```
mangelajo, line 63: ahh nice so you can get the issuer from the exiting token :D

NickCao: That's the best part about OIDC, everything could then be deducted from that one url.

packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -0,0 +1,89 @@
...
+
+import aiohttp
+import asyncclick as click
+import truststore
+from aiohttp import web
+from anyio import create_memory_object_stream
+from anyio.to_thread import run_sync
+from authlib.integrations.requests_client import OAuth2Session
+from joserfc.jws import extract_compact
+from yarl import URL
```
mangelajo, line 13: Can we avoid this extra dependency?, I just saw it sorry.

NickCao: yarl is already a dependency of aiohttp, not an extra.

packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -0,0 +1,89 @@
...
+from typing import ClassVar
+
+import aiohttp
+import asyncclick as click
+import truststore
+from aiohttp import web
+from anyio import create_memory_object_stream
+from anyio.to_thread import run_sync
+from authlib.integrations.requests_client import OAuth2Session
+from joserfc.jws import extract_compact
```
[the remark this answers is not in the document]
NickCao: We could use authlib.jose: https://docs.authlib.org/en/latest/jose/index.html, but it's in the process to be split into joserfc.

mangelajo (approved):

PR #283, opened 2025-02-17
client: add default handler

bennyz:
## Summary by CodeRabbit

- **New Features**
  - Introduced an automatic logging capability that ensures log messages are reliably captured.

### DIFF
--- a/packages/jumpstarter/jumpstarter/client/core.py
+++ b/packages/jumpstarter/jumpstarter/client/core.py
@@ -52,6 +52,12 @@ class AsyncDriverClient(
         self.logger = logging.getLogger(self.__class__.__name__)
         self.logger.setLevel(self.log_level)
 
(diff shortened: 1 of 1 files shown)


NickCao:
The logging handler should ideally be set at the "entrypoint" of the execution. If this is about using jumpstarter client as a library, it should be done in the `Lease` class.

bennyz:
> The logging handler should ideally be set at the "entrypoint" of the execution. If this is about using jumpstarter client as a library, it should be done in the `Lease` class.

do you mean where we instantiate the client? like https://github.com/jumpstarter-dev/jumpstarter/blob/19a0f821fdc71d9a13391acda53baeb04d4a8216/packages/jumpstarter/jumpstarter/client/client.py#L23

mangelajo (approved):

PR #293, opened 2025-02-19
Handle jmp-client exceptions and introduce a exception handling module

mangelajo:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new `client_lease` command in the CLI, replacing the previous `lease` command for a streamlined experience.

- **Bug Fixes / Improvements**
  - Enhanced error reporting across client operations for clearer, more actionable messages.
  - Improved error handling for gRPC-related exceptions, providing more specific feedback during connection issues.
  - Upgraded feedback mechanisms during lease and client configuration operations for better guidance.

### DIFF
--- a/packages/jumpstarter-cli-common/jumpstarter_cli_common/exceptions.py
+++ b/packages/jumpstarter-cli-common/jumpstarter_cli_common/exceptions.py
@@ -1,0 +1,36 @@
+import asyncclick as click
+
+from jumpstarter.common.exceptions import JumpstarterException
+
+
+class ClickExceptionRed(click.ClickException):
+    def format_message(self) -> str:
+        return click.style(self.message, fg='red')
+
+def async_handle_exceptions(func):
+    """Decorator to handle exceptions in async functions."""
+    async def wrapped(*args, **kwargs):
+        try:
+            return await func(*args, **kwargs)
+        except JumpstarterException as e:
+            raise ClickExceptionRed(str(e)) from None
+        except click.ClickException:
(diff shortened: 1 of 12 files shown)


packages/jumpstarter/jumpstarter/common/exceptions.py
```diff
@@ -0,0 +1,73 @@
...
+
+class FileAccessError(JumpstarterException):
+    """Raised when a file access error occurs."""
+    pass
+
+class FileNotFoundError(JumpstarterException):
+    """Raised when a file is not found."""
+    pass
+
+async def async_exception_handling(func):
```
mangelajo, line 48:
```suggestion
async def async_handle_exception(func):
```

packages/jumpstarter-cli-common/jumpstarter_cli_common/exceptions.py
```diff
@@ -0,0 +1,36 @@
...
+
+async def async_handle_exceptions(func):
+    """Decorator to handle exceptions in async functions."""
+    async def wrapped(*args, **kwargs):
+        try:
+            return await func(*args, **kwargs)
+        except JumpstarterException as e:
+            raise ClickExceptionRed(str(e)) from None
+        except click.ClickException:
+            raise # if it was already a click exception from the cli commands, just re-raise it
```
mangelajo, line 18: I am just being explicit here (as I could be leaving this to the "except Exception", but I want to make it clear that if we do the Exception later in time, we need to keep raising the ClickExceptions cleanly.

Issue #300, opened 2025-02-20
The j command throws an exception when an unknown command is used.


mangelajo:
```
jumpstarter ⚡local ➤ j abcd
Traceback (most recent call last):
  File "/home/fedora/jumpstarter/.venv/bin/j", line 10, in <module>
    sys.exit(j())
             ^^^
  File "/home/fedora/jumpstarter/packages/jumpstarter-cli-client/jumpstarter_cli_client/__init__.py", line 26, in j
    client.cli()(standalone_mode=False)
  File "/home/fedora/jumpstarter/.venv/lib/python3.12/site-packages/asyncclick/core.py", line 1211, in __call__
    return anyio.run(self._main, main, args, kwargs, **opts)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fedora/jumpstarter/.venv/lib/python3.12/site-packages/anyio/_core/_eventloop.py", line 74, in run
    return async_backend.run(func, args, {}, backend_options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fedora/jumpstarter/.venv/lib/python3.12/site-packages/anyio/_backends/_asyncio.py", line 2303, in run
    return runner.run(wrapper())
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/fedora/.local/share/uv/python/cpython-3.12.3-linux-x86_64-gnu/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fedora/.local/share/uv/python/cpython-3.12.3-linux-x86_64-gnu/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/fedora/jumpstarter/.venv/lib/python3.12/site-packages/anyio/_backends/_asyncio.py", line 2291, in wrapper
    return await func(*args)
           ^^^^^^^^^^^^^^^^^
  File "/home/fedora/jumpstarter/.venv/lib/python3.12/site-packages/asyncclick/core.py", line 1214, in _main
    return await main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #301, opened 2025-02-20
Avoid exceptions for unknown commands under j

mangelajo:
Without this change an unknown command results into a big exception, with standalone mode, a wrong command will result in a friendly reminder to look at the help output from click.

Fixes: #300 

Now the output looks like:
```
jumpstarter ⚡local ➤ j abcd
Usage: j [OPTIONS] COMMAND [ARGS]...
Try 'j --help' for help.

Error: No such command 'abcd'.
```

### DIFF
--- a/packages/jumpstarter-cli-client/jumpstarter_cli_client/__init__.py
+++ b/packages/jumpstarter-cli-client/jumpstarter_cli_client/__init__.py
@@ -23,7 +23,7 @@ def client(log_level: Optional[str]):
 
 def j():
     with env() as client:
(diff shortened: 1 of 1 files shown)


NickCao:
As I remembered the problem with standalone mode is that it would call sys.exit internally, causing the context managers to not be properly exited.

NickCao:
> As I remembered the problem with standalone mode is that it would call sys.exit internally, causing the context managers to not be properly exited.

It's us doing context managers wrong, they are properly exited.

NickCao (approved):

PR #302, opened 2025-02-20
Close client in the client_from_path context manager

NickCao:


NickCao:
Followup to https://github.com/jumpstarter-dev/jumpstarter/pull/301, ensures that clients are properly closed on sys.exit.

packages/jumpstarter-driver-network/jumpstarter_driver_network/adapters/fabric.py
```diff
@@ -1,30 +1,37 @@
...
-    connect_kwargs: dict[str, Any] | None = None
-    inline_ssh_env: bool | None = None
-
-    async def __aenter__(self):
-        addr = await super().__aenter__()
-        return Connection(
+@blocking
+@asynccontextmanager
+async def FabricAdapter(
+    *,
```
mangelajo, line 17: This is used to enforce all the others as keywords, right?

NickCao: Yes, as they were kw_only dataclasses. We might want to allow passing client as a positional argument later.

mangelajo (approved):
Looks good, a nice improvement.

I wonder if there would be a way of getting to the "next mile", doing it at the base client :D
```
fr>>> from jumpstarter.common.utils import serve
>>> client = serve(PySerial(url="loop://"))
>>> cl = client.open()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: '_GeneratorContextManager' object has no attribute 'open'
>>> cl = client.__enter__()
>>> p = cl.open()
>>> dir(p)
```

mangelajo:
oops it fails now, I guess it doesn,t work on top of the dbus without adapting the dbus now.

NickCao:
> oops it fails now, I guess it doesn,t work on top of the dbus without adapting the dbus now.

Indeed, fixing in a second.

NickCao:
> I wonder if there would be a way of getting to the "next mile", doing it at the base client :D

This would bring us back to the "hacky" realm, there has to be at least a toplevel context manager managing the ExitStack.

PR #305, opened 2025-02-21
MCU Workflow example with RP2350

mangelajo:


examples/mcu-rp2350/Containerfile.devspace
```diff
@@ -0,0 +1,9 @@
+FROM quay.io/jumpstarter-dev/jumpstarter-devspace:latest
+
+USER root
+RUN dnf install -y gcc systemd-devel gdb
+USER 10001
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```
NickCao, line 6: Or also use dnf as above?

mangelajo: ubi9 does not include rustup, just the packaged stuff.

fedora does but not sure if there is a devspaces developer base.

mangelajo:
I should update this based on the work here: 

https://gitlab.com/mangelajo/jumpstarter-kc25-devspace

PR #309, opened 2025-02-25
Provide a simplified "ref" access to proxy drivers

mangelajo:
This avoids over-verbose configuration, while maintaining the simplicity of a Proxy driver still provides a better administrator/user experience.

The Exporter config example now includes this and provides a more elaborate example.

## Summary by CodeRabbit

- **Documentation**
  - Reorganized driver documentation for improved navigation, including updated ordering and clearer explanations for the Proxy driver.
- **New Features**
  - Introduced a streamlined configuration approach using dot-delimited driver paths.
  - Expanded Exporter settings with additional driver options, including `power`, `serial`, `storage`, and `custom` sections.
  - Added a new optional `ref` attribute in Exporter configuration to simplify driver referencing.
  - Added a new entry point for the Proxy driver in the project configuration.

### DIFF
--- a/docs/source/api-reference/drivers/proxy.yaml
+++ b/docs/source/api-reference/drivers/proxy.yaml
@@ -2,7 +2,9 @@ children:
   proxy:
     type: "jumpstarter_driver_composite.driver.Proxy"
     config:
-      path: [foo, bar, power] # full path of the underlying driver
+      path: "foo.bar.power" # full path of the underlying driver
+  refproxy: # simplified syntax for Proxy
+    ref: "foo.bar.power"
   foo:
     children:
       bar:
--- a/docs/source/api-reference/drivers/proxy.yaml
+++ b/docs/source/api-reference/drivers/proxy.yaml
@@ -1,9 +1,5 @@
 children:
   proxy:
-    type: "jumpstarter_driver_composite.driver.Proxy"
-    config:
-      path: "foo.bar.power" # full path of the underlying driver
-  refproxy: # simplified syntax for Proxy
     ref: "foo.bar.power"
   foo:
     children:
--- a/packages/jumpstarter/jumpstarter/config/exporter_test.py
+++ b/packages/jumpstarter/jumpstarter/config/exporter_test.py
@@ -120,7 +120,6 @@ export:
         export={
             "power": ExporterConfigV1Alpha1DriverInstance(
                 type="jumpstarter_driver_power.driver.PduPower",
-                children={},  # missing children defaults to empty
                 config={
                     "host": "192.168.1.111",
                     "port": 1234,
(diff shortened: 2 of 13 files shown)


docs/source/api-reference/drivers/proxy.yaml
```diff
@@ -2,7 +2,9 @@ children:
   proxy:
     type: "jumpstarter_driver_composite.driver.Proxy"
     config:
-      path: [foo, bar, power] # full path of the underlying driver
+      path: "foo.bar.power" # full path of the underlying driver
+  refproxy: # simplified syntax for Proxy
+    ref: "foo.bar.power"
```
mangelajo, line 7: Still doctest is failing with: 

@NickCao any idea of why could this be happening?

```
    pydantic_core._pydantic_core.ValidationError: 1 validation error for Proxy
    path
      Input should be a valid string [type=string_type, input_value=['foo', 'bar', 'power'], input_type=list]
        For further information visit https://errors.pydantic.dev/2.10/v/string_type

```

docs/source/api-reference/drivers/proxy.yaml
```diff
@@ -1,11 +1,6 @@
...
-    children:
-      bar:
-        children:
-          power:
-            type: "jumpstarter_driver_power.driver.MockPower"
+proxy:
+  ref: "foo.bar.power"
+foo:
+  bar:
+    power:
```
mangelajo, line 5: very cool, my only worry would be: do we expect to want any other reserved works (like type, config, ref) ? now it may be a good moment to reserve something like that to avoid conflict on existing configurations.

packages/jumpstarter/jumpstarter/config/exporter_test.py
```diff
@@ -26,15 +26,13 @@ async def test_exporter_serve(mock_controller):
                 type="jumpstarter_driver_power.driver.MockPower",
             ),
             "nested": ExporterConfigV1Alpha1DriverInstance(
-                children={
```
mangelajo, line 29: I would continue to test and validate that "children" works.

May be we want to keep the "non-children" method experimental until we figure out if we need more reserved keys?

packages/jumpstarter/jumpstarter/config/exporter_test.py
```diff
@@ -96,11 +94,10 @@ def test_exporter_config(monkeypatch: pytest.MonkeyPatch, tmp_path: Path):
       port: "/dev/ttyUSB0"
       baudrate: 115200
   nested:
-    children:
-      custom:
-        type: "vendorpackage.CustomDriver"
-        config:
-          hello: "world"
+    custom:
```
mangelajo, line 97: same, if we keep the non-children functionality let's test both.

NickCao:
Let's move the last commit into a separate PR, the ref part is good to go.

packages/jumpstarter/jumpstarter/config/exporter_test.py
NickCao: pydantic validates union types in a "smart" way in the sense that it would try all the options one by one and find the best match, so the "reserved" keys aren't really reserved, you can use them as usual and pydantic would figure it out. Yes, this is ambiguous and confusing, but it does work that way...

mangelajo:
Lol, it does not let me approve because I am also partly author. @bennyz we need your help! :D

mangelajo:
+1 on my side %)

NickCao (approved):

PR #311, opened 2025-02-26
Use Opendal driver in Http driver

NickCao:
## Summary by CodeRabbit

- **Refactor**
  - Streamlined the HTTP components by removing direct file management actions to focus solely on server control.
- **Tests**
  - Simplified the testing setup to ensure a more direct and reliable interaction with the server and enhanced test coverage for file operations.
- **Chore**
  - Introduced new dependencies to support enhanced composite storage functionality.
- **New Features**
  - Added methods for writing and reading byte data directly to and from remote files.
- **Documentation**
  - Updated API reference to reflect changes in file writing methods and added assertions for data integrity.

### DIFF
--- a/packages/jumpstarter-driver-http/jumpstarter_driver_http/driver_test.py
+++ b/packages/jumpstarter-driver-http/jumpstarter_driver_http/driver_test.py
@@ -1,14 +1,8 @@
-import os
-import uuid
-from tempfile import TemporaryDirectory
-
 import aiohttp
-import anyio
 import pytest
-from anyio import create_memory_object_stream
 
 from .driver import HttpServer
-from jumpstarter.common.resources import ClientStreamResource
+from jumpstarter.common.utils import serve
 
 
 @pytest.fixture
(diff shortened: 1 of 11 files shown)


packages/jumpstarter-driver-http/jumpstarter_driver_http/driver_test.py
```diff
@@ -17,44 +11,32 @@ def anyio_backend():
...
+    file.write(str(tmp_path / local_filename))
+    file.close()
 
-    async with anyio.create_task_group() as tg:
-        tg.start_soon(send_data)
+    print(http.storage.stat(filename))
 
-        uploaded_url = await server.put_file(filename, resource_handle)
-        assert uploaded_url == f"{server.get_url()}/{filename}"
+    uploaded_url = f"{http.get_url()}/{filename}"
```
mangelajo, line 37: Can we still keep a helper that handles the put_file and returns the url?

packages/jumpstarter-driver-http/jumpstarter_driver_http/driver_test.py
```diff
@@ -17,44 +11,32 @@ def anyio_backend():
...
-    server.resources[resource_uuid] = receive_stream
-
-    resource_handle = ClientStreamResource(uuid=resource_uuid).model_dump(mode="json")
+    (tmp_path / local_filename).write_bytes(test_content)
 
-    async def send_data():
-        await send_stream.send(test_content)
-        await send_stream.aclose()
+    file = http.storage.open(filename, "wb")
+    file.write(str(tmp_path / local_filename))
```
mangelajo, line 32: isn't this writing the path instead of test content?  Where does the "/" operator come from?

mangelajo (commented):
some questions.

packages/jumpstarter-driver-http/jumpstarter_driver_http/driver_test.py
NickCao: We should rename the `write` method then... It is writing the file content from the path. The `/` operator is from https://docs.python.org/3/library/pathlib.html#operators

bennyz (approved):

PR #319, opened 2025-02-28
Enable dynamic USB ports

mangelajo:
This is useful when dealing with devices which will create usb serial ports dynamically, and while you still want to interact with  those ports, the ports won't be available when you start the exporter.


mangelajo:
@NickCao this is a variation of something you proposed.

And I believe that while we want to check in some cases the device is there... we don't need the .device attribute anymore?.. let's see what tests dictate...

NickCao (approved):

PR #334, opened 2025-03-07
Install trurl into utils image

NickCao:
## Summary by CodeRabbit

- **Chores**
  - Updated the system setup process to include an additional dependency for enhanced functionality.


NickCao:
> Can we check if trurl is available in ubi9 images?

Sadly both of them are not, but I'm sure we can find replacements.

mangelajo (approved):
ok, let's live with it for now, we must be aware that we will need to find replacements, may be easy rsa just becomes the integration to cert-manager.

PR #356, opened 2025-03-18
treewide: improved handling of storage device

NickCao:
## Summary by CodeRabbit

- **Refactor**
	- Streamlined the writing and reading processes for storage devices by removing unnecessary checks and utilizing new utility functions.
	- Introduced robust error handling and logging for storage operations in the new storage module.
- **New Features**
	- Added new asynchronous functions for managing storage devices, including checking availability, writing, and reading data.
	- Enhanced configuration options for storage operations with new timeout parameters.

### DIFF
--- a/packages/jumpstarter-driver-sdwire/jumpstarter_driver_sdwire/driver.py
+++ b/packages/jumpstarter-driver-sdwire/jumpstarter_driver_sdwire/driver.py
@@ -18,6 +18,9 @@ class SDWire(StorageMuxFlasherInterface, Driver):
     itf: usb.core.Interface = field(init=False)
 
     storage_device: str | None = field(default=None)
+    storage_timeout: int = field(default=10)
+    storage_leeway: int = field(default=6)
+    storage_fsync_timeout: int = field(default=900)
 
     def effective_storage_device(self):
         if self.storage_device is None:
@@ -103,10 +106,22 @@ class SDWire(StorageMuxFlasherInterface, Driver):
     async def write(self, src: str):
         self.host()
         async with self.resource(src) as res:
-            await write_to_storage_device(self.effective_storage_device(), res, logger=self.logger)
+            await write_to_storage_device(
+                self.effective_storage_device(),
+                res,
+                timeout=self.storage_timeout,
+                leeway=self.storage_leeway,
+                fsync_timeout=self.storage_fsync_timeout,
+                logger=self.logger,
+            )
 
     @export
     async def read(self, dst: str):
         self.host()
         async with self.resource(dst) as res:
-            await read_from_storage_device(self.effective_storage_device(), res, logger=self.logger)
+            await read_from_storage_device(
+                self.effective_storage_device(),
+                res,
+                timeout=self.storage_timeout,
+                logger=self.logger,
+            )
--- a/packages/jumpstarter-driver-sdwire/jumpstarter_driver_sdwire/driver.py
+++ b/packages/jumpstarter-driver-sdwire/jumpstarter_driver_sdwire/driver.py
@@ -1,15 +1,13 @@
 from __future__ import annotations
 
-import os
 from dataclasses import dataclass, field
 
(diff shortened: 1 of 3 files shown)


packages/jumpstarter-driver-sdwire/jumpstarter_driver_sdwire/driver.py
```diff
@@ -130,10 +131,12 @@ async def wait_for_storage_device(self):
     async def write(self, src: str):
         self.host()
         storage_device = await self.wait_for_storage_device()
-        async with await FileWriteStream.from_path(storage_device) as stream:
+        file = open(storage_device, "wb")
+        async with FileWriteStream(file) as stream:
             async with self.resource(src) as res:
                 async for chunk in res:
                     await stream.send(chunk)
+            os.fsync(file.fileno())
```
mangelajo, line 139: An alternative could be opening with O_DIRECT I believe

 fd = os.open(file_path, os.O_WRONLY | os.O_CREAT | os.O_DIRECT, 0o644)
    
async with FileWriteStream(os.fdopen(fd, 'wb', buffering=0)) as stream:
        async with src as res:
            async for chunk in res:
                # Ensure chunk is aligned to BLOCK_SIZE
                if len(chunk) % BLOCK_SIZE != 0:
                    padding = BLOCK_SIZE - (len(chunk) % BLOCK_SIZE)
                    chunk += b'\x00' * padding  # Zero-padding
                await stream.send(chunk)
                
   but on MacOS you can't use O_DIRECT you would need to do a fcntl.fcntl(fd, fcntl.F_NOCACHE, 1)  I believe

The only advantage here would be that we would have better control of how much has been really written, but... anyway it would probably be the stream reception also buffering....

let's stick to the simpler fsync and let's see how it goes...

mangelajo: Another comment, is that on the previous generation of jumpstarter I was doing this, and I found it necessary to also sleep for a while after the sync: reason-> the disk will also have some buffering, and if you power it off too quicky the last blocks won't be written.

NickCao: `O_DIRECT` doesn't really work with anyio, and is not universally supported.

packages/jumpstarter/jumpstarter/common/storage.py
```diff
@@ -0,0 +1,115 @@
...
+                                total_bytes / (1024 * 1024),
+                                storage_device,
+                            )
+                        )
+                        next_print += 50 * 1024 * 1024
+
+            with fail_after(timeout):
+                while True:
+                    try:
+                        os.fsync(file.fileno())
```
mangelajo, line 86: The fsync time is really hard to predict, It's really system/hardware dependent.

For example, a system with a lot of RAM will buffer almost everything...
and then fsync can take minutes (5-10m) if it's a slow storage (like SD card).

May be leave like a really high timeout (15 min) to be safe. As long as it's user overrideable in config.

packages/jumpstarter-driver-sdwire/jumpstarter_driver_sdwire/driver.py
```diff
@@ -101,45 +99,14 @@ def dut(self):
...
     @export
     async def write(self, src: str):
         self.host()
-        storage_device = await self.wait_for_storage_device()
-        async with await FileWriteStream.from_path(storage_device) as stream:
-            async with self.resource(src) as res:
-                async for chunk in res:
-                    await stream.send(chunk)
+        async with self.resource(src) as res:
+            await write_to_storage_device(self.effective_storage_device(), res, logger=self.logger)
```
mangelajo, line 106: leeway and fsync timeouts should be configurable on the driver (here + dutlink)

mangelajo (approved):
Nice, We need to document the new config fields in the driver docs.

PR #361, opened 2025-03-19
[WIP] direct client to exporter connectivity

NickCao:


packages/jumpstarter/jumpstarter/common/utils.py
```diff
@@ -80,7 +89,13 @@ def env():
...
 
 
-def launch_shell(host: str, context: str, allow: list[str], unsafe: bool) -> int:
+def launch_shell(
+    host: str,
+    context: str,
+    allow: list[str],
+    unsafe: bool,
+    use_alternative_endpoints: bool,
+) -> int:
```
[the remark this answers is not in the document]
kirkbrauer: +1

packages/jumpstarter/jumpstarter/common/utils.py
```diff
@@ -98,6 +113,7 @@ def launch_shell(host: str, context: str, allow: list[str], unsafe: bool) -> int
         | {
             JUMPSTARTER_HOST: host,
             JMP_DRIVERS_ALLOW: "UNSAFE" if unsafe else ",".join(allow),
+            JMP_USE_ALTERNATIVE_ENDPOINTS: "1" if use_alternative_endpoints else "0",
```
kirkbrauer, line 116: Suggest changing to `JMP_USE_ALT_ENDPOINTS` to reduce the length of the environment variables.

packages/jumpstarter/jumpstarter/config/client.py
```diff
@@ -50,6 +50,8 @@ class ClientConfigV1Alpha1(BaseModel):
     token: str
     grpcOptions: dict[str, str | int] | None = Field(default_factory=dict)
 
+    useAlternativeEndpoints: bool = False
```
kirkbrauer, line 53: Might want to consider using field alias to prevent having to use non-pythonish keys here.

packages/jumpstarter/jumpstarter/exporter/tls.py
```diff
@@ -0,0 +1,98 @@
...
+    host, sep, port = endpoint.rpartition(":")
+
+    if sep == "":
+        raise ValueError("port not specified in endpoint {}".format(endpoint))
+
+    host = host.strip("[]")  # strip brackets from ipv6 addresses
+
+    try:
+        port = int(port)
+        if port < 0 or port > 65535:
```
kirkbrauer, line 22: Add consts for hardcoded values.

packages/jumpstarter/jumpstarter/exporter/tls.py
```diff
@@ -0,0 +1,98 @@
...
+
+import grpc
+from cryptography import x509
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import hashes, serialization
+from cryptography.hazmat.primitives.asymmetric import rsa
+from jumpstarter_protocol import jumpstarter_pb2
+
+
+def parse_endpoint(endpoint):
```
kirkbrauer, line 12: Maybe we can refactor this to use a standard library to parse the endpoint to prevent any possible issues with this code.

NickCao: In fact we don't even have to parse the endpoint for SAN here, we can use a certificate with arbitrary SAN and skip SAN verification on the client side, since the certificate is pinned anyway.

packages/jumpstarter/jumpstarter/exporter/tls.py
```diff
@@ -0,0 +1,98 @@
...
+    except ValueError as e:
+        raise ValueError("invalid port {} in endpoint {}".format(port, endpoint)) from e
+
+    try:
+        return ip_address(host), port
+    except ValueError:
+        return host, port
+
+
+def with_alternative_endpoints(server, endpoints: list[str]):
```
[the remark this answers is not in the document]
kirkbrauer: @NickCao @mangelajo is this doing peer-to-peer mTLS?

packages/jumpstarter/jumpstarter/config/client.py
kirkbrauer: Take a look at the `jumpstarter_kubernetes` package, I have re-aligned all of the CRDs in there to follow this convention of snake_case names in Python, but camelCase in the output YAML/JSON.

packages/jumpstarter/jumpstarter/exporter/tls.py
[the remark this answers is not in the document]
NickCao: Yes

packages/jumpstarter/jumpstarter/exporter/tls.py
NickCao: Dropped the entire parse_endpoint dance in favor of a dummy san.

PR #374, opened 2025-03-24
Reduce max depth to 4 on cli doc index

NickCao:


raballew:
Obsolete since #391 has been merged.

PR #377, opened 2025-03-25
Speedup tests with pytest-xdist

NickCao:
## Summary by CodeRabbit

- **Tests**
  - Enhanced the testing framework to support parallel execution for faster, more efficient test runs.
  - Optimized configuration to automatically utilize available system resources during testing.


mangelajo:
As talking with @NickCao on 1:1 it's really not providing any improvement, due to the fact that QEMU/arm emulation is so slow..

PR #386, opened 2025-03-31
Dedup output mode handling

NickCao:
## Summary by CodeRabbit

- **New Features**
	- Introduced a unified serialization framework standardizing output formats (JSON, YAML, NAME, PATH) across CLI commands for consistent user experience.
	- Added dynamic output option generation based on available serialization methods of resource types.
	- Added a new base model class encapsulating common serialization and naming logic for various resources.

- **Refactor**
	- Streamlined output handling in create, update, get, and print commands by consolidating format-specific operations into a unified `dump` method.
	- Updated version, configuration, and Kubernetes resource models to leverage the new serialization layer, removing redundant serialization methods.
	- Simplified output decorators to use type-parameterized automatic output options for better type-driven output handling.
	- Refactored client, exporter, and lease classes to inherit from common base classes, unifying serialization and naming logic.
	- Removed deprecated list container class and replaced with a generic serializable list model.
	- Introduced a new `echo` function to control output without automatic newlines, improving output formatting consistency.
	- Enhanced error handling for empty resource lists when no output format is specified.


kirkbrauer:
Looking good! @NickCao, let's also move the `V1Alpha1List` out of the `jumpstarter-kubernetes` package as it should go into the core now that it's going to be used more generally by the `jmpstarter-cli-pkg` package.

kirkbrauer:
@NickCao maybe we want to attack the configs in a different PR, but I think we should handle them with the same logic with added capabilities for file I/O and ordering.

kirkbrauer:
@NickCao I would love if we could try to push this through, would like to get it merged before we go to merge #382

NickCao:
> Looking good! @NickCao, let's also move the `V1Alpha1List` out of the `jumpstarter-kubernetes` package as it should go into the core now that it's going to be used more generally by the `jmpstarter-cli-pkg` package.

But it's still something kubernetes specific? The grpc lists are of a different structure.

NickCao:
> @NickCao maybe we want to attack the configs in a different PR, but I think we should handle them with the same logic with added capabilities for file I/O and ordering.

Surely yes, but currently their `dump_yaml` method has a slightly different signature.

PR #390, opened 2025-04-04
Support specifying the default content of partitions in qemu driver

NickCao:
## Summary by CodeRabbit

- **New Features**
	- Enhanced partition validation with an optional fallback to default partition settings when a specified partition isn’t found.
- **Tests**
	- Updated test configurations to incorporate the new default partition mappings and removed redundant operations, streamlining the process.

### DIFF
--- a/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
+++ b/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
@@ -44,7 +44,9 @@ class QemuFlasher(FlasherInterface, Driver):
 
     @export
     async def dump(self, target, partition: str | None = None):
-        async with await FileReadStream.from_path(self.parent.validate_partition(partition)) as stream:
+        async with await FileReadStream.from_path(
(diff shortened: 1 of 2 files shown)


packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -229,6 +231,8 @@ class Qemu(Driver):
     username: str = "jumpstarter"
     password: str = "password"
 
+    default_partitions: dict[str, Path] = field(default_factory=dict)
```
mangelajo, line 234: I would pre-initialize this with the know system paths for package installs

mangelajo (commented):
Looks great, An idea for follow ups

mangelajo (approved):

packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
NickCao: > I would pre-initialize this with the know system paths for package installs

This logic is present in the automotive image runner, we can copy that.

PR #435, opened 2025-04-30
Init hatch-pin-jumpstarter hook

NickCao:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new build plugin that automatically pins certain dependencies' versions during the build process.
- **Chores**
  - Added a new package for the build plugin and integrated it into the build configuration across all relevant packages.
  - Updated workspace configuration to include the new build plugin package.
  - Modified test commands to handle specific exit codes gracefully.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ docs-linkcheck:
 	uv run --isolated --all-packages --group docs $(MAKE) -C docs linkcheck
 
 pkg-test-%: packages/%
-	uv run --isolated --directory $< pytest
+	uv run --isolated --directory $< pytest || [ $$? -eq 5 ]
 
 pkg-mypy-%: packages/%
 	uv run --isolated --directory $< mypy .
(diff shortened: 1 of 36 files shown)


Makefile
```diff
@@ -24,7 +24,7 @@ docs-linkcheck:
 	uv run --isolated --all-packages --group docs $(MAKE) -C docs linkcheck
 
 pkg-test-%: packages/%
-	uv run --isolated --directory $< pytest
+	uv run --isolated --directory $< pytest || [ $$? -eq 5 ]
```
mangelajo, line 27: Why do we get retcode 5 sometimes? let's document (comment) the why we ignore it, is it safe?

mangelajo: Ok, it's the hatch-pin-jumpstarter package having no test. I will follow up with adding some test, even if noop

mangelajo (approved):

PR #437, opened 2025-05-05
Android Emulator Driver

kirkbrauer:
This PR includes a full implementation of an Android Driver providing emulation through the Android emulator CLI, debugging through [`adb`](https://developer.android.com/tools/adb), and remote screen access through the [`scrcpy`](https://github.com/Genymobile/scrcpy) tool.

## Drivers
- `AndroidEmulator` - Top-level driver for running Android emulators.
- `AndroidEmulatorPower` - Power driver implementation for stopping/starting android emulators.
- `AdbServer` - ADB server implementation to start/stop the ADB daemon in the background and port-forward ADB over Jumpstarter.
- `Scrcpy` - Scrcpy driver based on the `AdbServer` to provide a "stub" for the Scrcpy client to connect to.

## Clients
- `AndroidClient` - Composite client for Android devices/emulators.
- `AdbClient` - Python and CLI ADB client that provides a thin wrapper around the `adb` command with automatic port-forwarding and supports the [`adbutils`](https://github.com/openatx/adbutils) Python package for pure Python implementation.
- `ScrcpyClient` - CLI Scrcpy client that provides a thin wrapper around `scrcpy` to enable automatic port-forwarding on the client side. Requires the `scrcpy` CLI tool.

## Usage
```
~/repos/jumpstarter ⚡local ➤ j android
Usage: j android [OPTIONS] COMMAND [ARGS]...

  Generic composite device

Options:
  --help  Show this message and exit.

Commands:
  adb     Run adb using a local executable against the remote adb server.
  power   Generic power
  scrcpy  Run scrcpy using a local executable against the remote adb server.
```

## Example
```yaml
# local-emulator.yaml

apiVersion: jumpstarter.dev/v1alpha1
kind: ExporterConfig
metadata:
  name: local-exporter
  namespace: default
endpoint: ""
token: ""
export:
  android:
    type: jumpstarter_driver_android.driver.AndroidEmulator
    config:
      emulator:
        emulator_path: /Users/jdoe/Library/Android/sdk/emulator/emulator
        avd: "Pixel_9_Pro"
```

```bash
$ jmp shell --exporter-config ./local-emulator.yaml
~/repos/jumpstarter ⚡local ➤ j android power on
~/repos/jumpstarter ⚡local ➤ j android adb devices
List of devices attached
emulator-5554   device
~/repos/jumpstarter ⚡local ➤ j android scrcpy 
```


mangelajo (commented):
cool cool cool :)

docs/source/reference/package-apis/drivers/index.md
```diff
@@ -56,6 +58,16 @@ Drivers that handle media streams:
 * **[UStreamer](ustreamer.md)** (`jumpstarter-driver-ustreamer`) - Video
   streaming functionality
 
+### Virtualization Drivers
+
+Drivers for running virtual machines and systems:
```
NickCao, line 63: `running or connecting to`?

packages/jumpstarter-driver-android/jumpstarter_driver_android/client.py
```diff
@@ -0,0 +1,235 @@
...
+    pass
+
+
+class AdbClientBase(DriverClient):
+    """
+    Base class for ADB clients. This class provides a context manager to
+    create an ADB client and forward the ADB server address and port.
+    """
+
+    def _check_port_in_use(self, host: str, port: int) -> bool:
```
NickCao, line 30: This function is unused.

packages/jumpstarter-driver-android/jumpstarter_driver_android/client.py
```diff
@@ -0,0 +1,235 @@
...
+                return True
+        finally:
+            sock.close()
+        return False
+
+    @contextmanager
+    def forward_adb(self, host: str, port: int) -> Generator[str, None, None]:
+        """
+        Port-forward remote ADB server to local host and port.
+        If the port is already bound, yields the existing address instead.
```
NickCao, line 46: Not yet implemented?

packages/jumpstarter-driver-android/jumpstarter_driver_android/driver/device.py
```diff
@@ -0,0 +1,35 @@
...
+    A base Android device driver composed of the `AdbServer` and `Scrcpy` drivers.
+    """
+
+    @classmethod
+    def client(cls) -> str:
+        return "jumpstarter_driver_android.client.AndroidClient"
+
+    adb: AdbOptions = field(default_factory=AdbOptions)
+    disable_scrcpy: bool = field(default=False)
+    disable_adb: bool = field(default=False)
```
NickCao, line 24: What about making them enable options and True by default? Double negative can get confusing.

packages/jumpstarter-driver-android/jumpstarter_driver_android/client.py
kirkbrauer: This should also be removed after some experimentation, not reliable across platforms.

packages/jumpstarter-driver-android/jumpstarter_driver_android/driver/device.py
kirkbrauer: Agreed, will change this, maybe we should disable scrcpy by default, it can cause performance impacts on some virtual targets which is not idea. We should really be using VNC through QEMU by default, but I did not have time to implement this.

mangelajo:
> @mangelajo I'm not sure if we should shelve this in favor of a Cuttlefish-based driver or try to merge what we have now. I don't really want to include the entire Android SDK in a container image, so maybe it's best to wait for Cuttlefish.

I see you concern, I don't know, even with the known limitations, should we merge it anyway? , then we can make a better version later in time with cuttlefish?

It was awesome work, would probably attract people to jumpstarter IMHO!

PR #449, opened 2025-05-06
Use regular (non async) click

NickCao:


mangelajo:
It looks good, we should probably continue after https://github.com/jumpstarter-dev/jumpstarter/pull/443 and document our async/blocking strategy, explain where developers should use async or blocking, and why.

mangelajo:
We should also provide some docstrings on those wrappers, so future developers know what each wrapper exactly does without looking at the code itself.

mangelajo (approved):
Approved with comments to improve.

NickCao:
Checked that the cli reference doc still renders correctly, will followup by moving the decorator to jumpstater core, add docstring and make use of it in more places.

PR #451, opened 2025-05-07
 Use rich logging handler

NickCao:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new command-line option to set the logging level for CLI commands, providing enhanced control over log output.

- **Chores**
  - Updated dependencies to add or remove the "rich" library as needed for logging improvements.

- **Refactor**
  - Streamlined logging configuration by centralizing log-level handling and removing redundant logging setup from individual CLI commands.

### DIFF
--- a/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/__init__.py
+++ b/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/__init__.py
@@ -1,5 +1,8 @@
+import logging
+
 import click
 from jumpstarter_cli_common.alias import AliasedGroup
+from jumpstarter_cli_common.opt import opt_log_level
 from jumpstarter_cli_common.version import version
 
 from .create import create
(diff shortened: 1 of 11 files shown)


packages/jumpstarter-cli-admin/jumpstarter_cli_admin/__init__.py
```diff
@@ -10,6 +11,7 @@
 
 
 @click.group(cls=AliasedGroup)
+@opt_log_level
```
mangelajo, line 14: Wasn't the problem with this that then the logging option only appears way back in the cmdline ? instead of the last level command?

I believe we wanted to get away from that. I will test empirically and report back.

mangelajo: well, this just moves it to "jmp admin --log-level"

instead of jmp admin create --log-level or the other commands.


IMHO your proposal is better, because we cover more "range" from a single place., the ideal would be to have this at the base on every command, but not sure if that is something we can accomplish cleanly with click

mangelajo (approved):

Issue #455, opened 2025-05-08
[TRACKER] TLS insecure is the default in configuration


mangelajo:
The jumpstarter-controller endpoints are secured by TLS, but in 0.6 the certificates
are rotated every time the controller is restarted. This means that the certificates are
not trusted by default. 

Because of this we keep tls.insecure defaulting to true.

This should be fixed for 0.7, by allowing proper handling of certificates in the jumpstarter-controller.

PR #456, opened 2025-05-08
Enable tls.insecure true by default

mangelajo:
Our jumpstarter-controller currently re-creates the TLS certificates on every restart. This means that clients and exporters cannot trust the gRPC certificates by default.

This commit sets tls insecure by default to avoid user frustration, this should be fixed for 0.7 once the controller supports proper certificate handling.

Related-Issue: #455

## Summary by CodeRabbit

- **Documentation**
  - Added warnings in configuration and installation guides about temporary TLS certificate rotation and trust issues in version 0.6.
  - Updated example configurations to set `tls.insecure: true` by default and included guidance on disabling TLS verification.
- **New Features**
  - Default client and exporter configurations now disable TLS verification to accommodate rotated, untrusted certificates in version 0.6.

### DIFF
--- a/docs/source/getting-started/configuration/files.md
+++ b/docs/source/getting-started/configuration/files.md
@@ -29,7 +29,17 @@ config:
 **File**: Various files with `.yaml` extension  
 **Location**: `~/.config/jumpstarter/clients/*.yaml`  
 **Description**: Stores client configurations including endpoints, access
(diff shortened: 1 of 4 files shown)


mangelajo:
Closing in favor of https://github.com/jumpstarter-dev/jumpstarter/pull/462

PR #506, opened 2025-05-28
add support for RcarS4

bennyz:
Signed-off-by: Benny Zlotnik <bzlotnik@redhat.com>


.gitattributes
```diff
@@ -0,0 +1 @@
+packages/jumpstarter-driver-flashers/oci_bundles/rcar_s4/data/ filter=lfs diff=lfs merge=lfs -text
```
bennyz, line 1: need to remember to remove the artifacts

michalskrivanek: I would prefer to include the scripts used to build them. we don't need to automate that but having that available would be helpful.

mangelajo: Me too, either instructions to download them (links, or just a curl script in the makefile), and for rootfs,  we could include them here somewhere. Even at some point we can build it in the github arm builders, but I would defer that for now.

mangelajo:
You need to add "click" to the pyproject of the rpi driver.

mangelajo:
@bennyz ^ otherwise it won't install the dependency and fail the unit tests.

mangelajo (approved):
We only need to change the image ref, and publish it.

packages/jumpstarter-driver-flashers/oci_bundles/rcar_s4/manifest.yaml
```diff
@@ -0,0 +1,19 @@
...
+  login:
+    login_prompt: "login:"
+    username: "root"
+    prompt: "#"
+  default_target: "emmc"
+  targets:
+    emmc: "/dev/mmcblk0"
+  kernel:
+    file: data/flasher.itb
+    address: "0x58000000"
```
michalskrivanek, line 19: nit: this is the address bootm uses by deafult, but if you change that (for other boards perhaps) then "bootm" without argumetns won't work. so maybe make it more explicit setting bootcmd to "bootm $0x58000000"

bennyz: I assume it's `bootm 0x58000000` or `bootm $loadddr`

mangelajo (approved):

PR #510, opened 2025-05-30
use FIT image for ti_j784s4 flasher

michalskrivanek:
using latest Fedora kernel.
It doesn't boot with bootm properly, perhaps due to some uboot misconfig, but we can still use the FIT format.

## Summary by CodeRabbit

- **New Features**
  - Added scripts and configuration files to enable building and flashing unified boot images for TI J784S4X EVM and AM69 platforms.
  - Introduced a script to wait for network interfaces to become active before startup.
  - Provided a new Buildroot configuration for streamlined kernel and root filesystem generation.
- **Improvements**
  - Simplified the boot process by switching to a single combined image format and updating boot commands accordingly.


packages/jumpstarter-driver-flashers/oci_bundles/ti_j784s4xevm/manifest.yaml
```diff
@@ -25,10 +25,5 @@ spec:
 #    - "dd if=/dev/zero of=/dev/mmcblk0 bs=512 count=34"
 #    - "dd if=/dev/zero of=/dev/mmcblk1 bs=512 count=34"
   kernel:
-    file: data/J784S4XEVM.flasher.img
```
mangelajo, line 28: https://github.com/jumpstarter-dev/jumpstarter/actions/runs/15352215504/job/43202958209?pr=510#step:8:400

Ok, so there is a problem with itb images.

We stop declaring the dtb section, which tells the flasher what the expected address is,
also I suspect that the current code does not support dtb to be empty.

The reason why we at least want that info and a separate dtb, is that we want to let customers switch the DTB
in realtime, for example, when they have a derivative of the board that they have evolved into product.

i.e. j flasher flash image.raw --dtb my-dtb.dtb

The flasher needs to know the address where to put the dtb provided by customer in that case.
Also we may want to let customers derive the OCI images and put their dtb in they want.


Can we provide an image with just the kernel + initrd , and then two separate .dtbs, one for
 J784S4EVM and another for SK-AM69? 

The other option is that we rewire the reflasher to fully understand fit, but it would still miss the ability to change the bundled dtb, kernel or ramdisk in "realtime".

michalskrivanek: empty dtb has been added by https://github.com/jumpstarter-dev/jumpstarter/pull/506/commits/f44dfec4c62aca654013aa1ba776a28220458367

I think it is quite flexible this way, the bundling allows us to "ship" dtbs that we anyway ship in kernel-automotive, but it also allows anything else to be loaded. each FIT component is extractable by imxtract and swapped with anything else

michalskrivanek:
the current version has initramfs embedded in the kernel, exceeding the default size uboot can handle. i'll try with separate initrd as well....

packages/jumpstarter-driver-flashers/oci_bundles/ti_j784s4xevm/manifest.yaml
mangelajo: Ah I see, yes 0x88000000 in this case.

Can we keep the dtb structure, with just the address? we may need to verify that the client/driver are fine with empty file, just address.

michalskrivanek:
keeping separate initrd the size is within the default CONFIG_SYS_BOOTM_LEN (64MB) so let's go with that, we can boot with a simple "bootm"

packages/jumpstarter-driver-flashers/oci_bundles/ti_j784s4xevm/manifest.yaml
michalskrivanek: it *may* work with am69 as well, supposedly it should try to load a compatible configuration. I havent' tried that. If it doesn't work the only change is to boot the right config with bootm 0x90000000#am69 or to supply another DTB externally

mangelajo:
This was finally implemented in https://github.com/jumpstarter-dev/jumpstarter/pull/535/files

PR #535, opened 2025-06-26
builroot-based flasher images

michalskrivanek:
A single config building two FIT images supporting similar-enough boards using either upstream or kernel-automotive kernel. Includes TI J784S4EVM, SK-AM69 and Renesas R-Car S4 Spider and SK.

## Summary by CodeRabbit

* **New Features**
  * Enhanced support for device tree blob (DTB) variants with customizable boot commands in manifests.
  * Added new manifests, kernel configurations, and scripts for TI J784S4 and Renesas R-Car S4 platforms.
  * Introduced automated FIT image building and kernel replacement scripts for aarch64 devices.
  * Added optional driver configuration parameters for selecting DTB variant and manifest file.

* **Bug Fixes**
  * Improved validation and error reporting for DTB variant selection and missing DTB files.

* **Documentation**
  * Updated README with new DTB variant and manifest configuration options.

* **Tests**
  * Added extensive tests covering boot command retrieval and DTB variant behavior.

* **Chores**
  * Updated bundle build scripts to include all YAML manifests and exclude placeholder files.
  * Added udev rules and network scripts for interface setup on Renesas platforms.


michalskrivanek:
[WIP - still needs the corresponding API changes for selecting the right config]

@bennyz @mangelajo this is a combination of #510 and #506 , using the single config building both kernel+initrd(6.14 kernel for TI) and kernel-automotive+initrd for S4

The configs are fully reused and the time to build is less as well as the second pass with kernel-automotive is a trivial repackage only (though note the initrds are slightly different due to the extra kernel-automotive modules)
If we go this way we can drop the TI and S4 specific bundles (and hopefully save ourselves from introducing new ones)

This produces two 70MB FITs, the complete bundle would be 140MB. It's not too bad, so I wonder if it's better to split this into two bundles or keep it all together with a single manifest and leave it up to the user to pick one or the other.

PR #537, opened 2025-07-01
Allow image building without root permissions

mangelajo:


mangelajo:
> breaks container builds on mac:/

Let's add automation to build from container then :)

i.e. a make target, container file or set of skopeo/buildah/podman calls in script perhaps?

michalskrivanek:
> > breaks container builds on mac:/
> 
> Let's add automation to build from container then :)
> 
> i.e. a make target, container file or set of skopeo/buildah/podman calls in script perhaps?

ok. how about #543 ?

Issue #563, opened 2025-07-14
bashrc disabled for `jmp shell`


vraiti:
It appears that `jmp shell` [forces no automatic loading of `~/.bashrc`](https://github.com/jumpstarter-dev/jumpstarter/blob/main/packages/jumpstarter/jumpstarter/common/utils.py). I have a few custom shell functions that I'd like to be able to use in the Jumpstarter shell, but it's a little inconvenient to have to `source ~/.bashrc` manually whenever I start a new shell. Could there be a way to configure this?

NickCao:
This is to prevent bashrc from overriding the PS1 set by jumpstarter, do you have any way of avoiding this?

vraiti:
https://github.com/jumpstarter-dev/jumpstarter/pull/566

PR #566, opened 2025-07-14
Use PROMPT_COMMAND to set jmp shell prompt

vraiti:
In reference to https://github.com/jumpstarter-dev/jumpstarter/issues/563


packages/jumpstarter/jumpstarter/common/utils.py
```diff
@@ -76,11 +76,11 @@ def launch_shell(
...
 
     if shell_name.endswith("bash"):
+        PS1=f"{ANSI_GRAY}{PROMPT_CWD} {ANSI_YELLOW}⚡{ANSI_WHITE}{context} {ANSI_YELLOW}➤{ANSI_RESET} "
         env = common_env | {
-            "PS1": f"{ANSI_GRAY}{PROMPT_CWD} {ANSI_YELLOW}⚡{ANSI_WHITE}{context} {ANSI_YELLOW}➤{ANSI_RESET} ",
+            'PROMPT_COMMAND': f'PS1="{PS1}"',
         }
-        cmd = [shell, "--norc", "--noprofile"]
-        process = Popen(cmd, stdin=sys.stdin, stdout=sys.stdout, stderr=sys.stderr, env=env)
+        process = Popen(shell, stdin=sys.stdin, stdout=sys.stdout, stderr=sys.stderr, env=env)
```
mangelajo, line 83:
```suggestion
        process = Popen([shell], stdin=sys.stdin, stdout=sys.stdout, stderr=sys.stderr, env=env)
```

Popen first arg is a list according to docs: https://docs.python.org/3/library/subprocess.html#popen-constructor

vraiti:
So turns out trying to set PROMPT_COMMAND leads to a whole host of issues since VTE will clobber it in /etc/profile.d/vte.sh. Instead I tried adding a user configuration shell_use_profiles (that is by default disabled). Lmk if this is ok.

vraiti:
@mangelajo

NickCao (commented):
Overall LGTM, but are the `'--norc','--noprofile'` portable across shells?

PR #570, opened 2025-07-16
flashers: fallback to request for metadata request

bennyz:
There is currently an issue with custom certificates in opendal, which fails with open/send requests. This patch falls back to using requests to fetch the metadata instead.


packages/jumpstarter-driver-flashers/jumpstarter_driver_flashers/client.py
```diff
@@ -299,14 +301,44 @@ def _sha256_file(self, src_operator, src_path) -> str:
...
+            metadata = src_operator.stat(src_path)
+            metadata_dict.update(
+                {
+                    "content_length": metadata.content_length,
+                    "etag": metadata.etag,
+                }
+            )
+        except Exception as e:
+            # TODO(bennyz): remove when opendal issue is sorted out
+            # https://github.com/apache/opendal/discussions/6418
```
mangelajo, line 321: 👍

PR #580, opened 2025-07-30
remove lgpio dependency

bennyz:
until container image can be built


michalskrivanek:
@bennyz no longer needed, with https://github.com/jumpstarter-dev/jumpstarter/pull/587 merged...

